PR #4, opened 2016-12-07
Standalone Solution

joelmartinez:
After the extraction of this project from the Mono repository, the project files need to be updated to reflect the fact that it does not live in the context of their makefile.


joelmartinez:
@jonpryor ... this is still very much a work in progress, but I wanted to get your input on how exactly I should proceed. So the ultimate goal is two-fold

- Have standalone solution/project files that let me just open, code, and debug (which was not possible before, at least for monodoc).
- _Also_ have solution/project files that will still work when we eventually re-integrate this into mono for distribution in those makefiles.

Now, the tricky part is that before, libraries like lucene.net were literally embedded into `monodoc.dll`, while other libraries like SharpZipLib, or Mono.Cecil were referenced from the locally built versions in the mono repo. Currently, I've gotten everything building by replacing all of those with nuget references and updating the csproj files so I can code in VS(fM).

So for point 1, that works, but I realize that won't exactly jive with point 2, since mono requires everything to be built via source for licensing reasons. Would love to get your thoughts on how we can proceed here :)

jonpryor:
First of all, there should be *one* solution file, not two, particularly since mdoc depends on monodoc *anyway*. Thus, there should be a top-level `api-doc-tools.sln`, not separate `mdoc/mdoc.sln` and `monodoc/monodoc.sln` files.

Let's call that "0". :-)

Secondly, much as I'd like to have -- and am *trying to have* -- an msbuild-only build system, that isn't *entirely* practical, at least for my experiences in `xamarin-android`.

Thus, I would suggest the following:

1. Ditch the NuGet packages. I know, they're *oh so tempting*, such that the `xamarin-android` repo used a NuGet package for Mono.Cecil. The problem is that this complicates licensing and source delivery -- though this is *presumably* addressable -- and more importantly, makes it *harder* to get fixes.

    Case in point: there were some known issues in Mono.Cecil 0.10.x, and the NuGet package hadn't been updated. On the assumption that there will be *further* discovered issues in Mono.Cecil, `xamarin-android` now uses a GitHub external repo reference. This allows `xamarin-android` to more easily update `Mono.Cecil` if (when!) bugs are found.

    Meaning the `api-doc-tools` repo should use external repo references, *not* NuGet packages.

2. You'll want a (minimal!) `Makefile`. It need not do *much*, but it needs to have "enough." In this case, "enough" means:

        git submodule update --init --recursive

    The `xamarin-android` repo places the above into a `make prepare` target, while the `make all` target delegates to MSBuild:

        MSBUILD = xbuild
        all:
            $(MSBUILD) api-doc-tools.sln

        prepare:
            git submodule update --init --recursive

    A jenkins build thus becomes:

        make prepare all

> Now, the tricky part is that before, libraries like lucene.net...

These would become external package references, just as `mono` imports them.

joelmartinez:
Thank you for the guidance, @jonpryor, I will reconfigure to a single solution, and add submodules for these other projects.

    MSBUILD = xbuild

Wouldn't this make it so that it only works when Mono is installed? In theory, mdoc and monodoc should build just fine with regular .net, right? or is there some other dependency on mono that I'm not thinking of?

jonpryor:
MSBUILD = xbuild

The idea would be to use a make variable to select between xbuild and msbuild.

Perhaps it should default to msbuild instead.

Regardless, using a make variable here allows the caller to switch between them, which might be useful/necessary when integrating with mono's build system:

    make MSBUILD=xbuild ...

jonpryor:
While you're working on this, you should also look into setting up a Jenkins lane to ensure that things build. :-)

Makefile
```diff
@@ -0,0 +1,27 @@
+MSBUILD = xbuild
+MDOC = mdoc/bin/Release/mdoc.exe
```
jonpryor: Opinions may vary, but I'd suggest a *toplevel* `bin` directory, not a *per-assembly* `bin` directory.

Maybe I'm the crazy one, but this is what I do in Java.Interop and xamarin-android; all build artifacts are placed in `$topdir/bin`.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+all: prepare build
```
jonpryor: `prepare` should *not* be an `all` dependency.

The rationale is that if I update a submodule in my local tree *without* a corresponding commit -- because I'm still writing code for that commit! -- `make all` *should not* **reset** the submodule I updated. That gets *really* annoying to deal with.

Java.Interop and xamarin-android have `make prepare` as separate targets, not part of `all`.

joelmartinez: ahh, duly noted

Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+clean:
+	$(MSBUILD) apidoctools.sln /t:clean
+	$(MSBUILD) external/SharpZipLib/ICSharpCode.SharpZipLib.NET45/ICSharpCode.SharpZipLib.csproj /t:clean
+	$(MSBUILD) external/Mono.Cecil/Mono.Cecil.sln /t:clean
+
+check: check-mdoc
+
+check-mdoc:
+	cd mdoc; $(MAKE) check
```
jonpryor: Should we just merge the contents of `mdoc/Makefile` into this file, and delete `mdoc/Makefile`?

joelmartinez: There's a lot of mdoc-specific tests in `mdoc/Makefile` (some of which I'm still working on making pass, am getting some subtly different results at the moment, not sure why)

joelmartinez:
@jonpryor ok, so I've got all the mdoc unit tests passing. Now I want to work on monodoc and I wanted to get some advice from you on how to proceed. 

The monodoc unit tests in mono are using some baked in functionality to run the tests with nunit on the command line. How would you recommend that we bring in that dependency?  Should I still bring that in via an external submodule, or do you think it might be ok to depend on nuget to [fetch it down](https://www.nuget.org/packages/NUnit.Console) (since we're not distributing it)?

also, I will be addressing the feedback you gave above, and also squashing all the commits once everything is ready to roll.

jonpryor:
>  run the tests with nunit on the command line.

I've been using [NUnit via NuGet package](https://github.com/xamarin/java.interop/blob/master/Makefile#L43).

Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+all: build
+
+build: $(MDOC)
+$(MDOC): external/SharpZipLib/bin/Release/ICSharpCode.SharpZipLib.dll external/Mono.Cecil/bin/net_4_0_Release/Mono.Cecil.dll
+	$(MSBUILD) apidoctools.sln /p:Configuration=Release
+
+prepare:
+	git submodule update --init --recursive
+	nuget restore -PackagesDirectory Packages
```
jonpryor: Why `Packages` and not `packages`? The typical directory appears to be `packages`, based on the number of them I have from `nuget restore`...

Differences in casing can cause issues on case-sensitive filesystems. We should be consistent.

Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+external/SharpZipLib/bin/Release/ICSharpCode.SharpZipLib.dll:
```
jonpryor: In keeping with a top-level `bin` directory, this should be:

    bin/ICSharpCode.SharpZipLib.dll:
        -mkdir -p "`dirname "$@"`"
        $(MSBUILD) external/SharpZipLib/ICSharpCode.SharpZipLib.NET45/ICSharpCode.SharpZipLib.csproj /p:Configuration=Release
        cp external/SharpZipLib/bin/Release/ICSharpCode.SharpZipLib.dll "$@"

Makefile
```diff
@@ -0,0 +1,31 @@
...
+	$(MSBUILD) external/SharpZipLib/ICSharpCode.SharpZipLib.NET45/ICSharpCode.SharpZipLib.csproj /p:Configuration=Release
+
+external/Mono.Cecil/bin/net_4_0_Release/Mono.Cecil.dll:
```
jonpryor: Ditto here

Makefile
```diff
@@ -0,0 +1,31 @@
+MSBUILD = xbuild
+MDOC = mdoc/bin/Release/mdoc.exe
+
+all: build
+
+build: $(MDOC)
+$(MDOC): external/SharpZipLib/bin/Release/ICSharpCode.SharpZipLib.dll external/Mono.Cecil/bin/net_4_0_Release/Mono.Cecil.dll
```
jonpryor: ...and here:

    $(MDOC): bin/ICSharpCode.SharpZipLib.dll bin/Mono.Cecil.dll

Makefile
```diff
@@ -0,0 +1,31 @@
...
+	git submodule update --init --recursive
+	nuget restore -PackagesDirectory Packages
+
+external/SharpZipLib/bin/Release/ICSharpCode.SharpZipLib.dll:
+	$(MSBUILD) external/SharpZipLib/ICSharpCode.SharpZipLib.NET45/ICSharpCode.SharpZipLib.csproj /p:Configuration=Release
+
+external/Mono.Cecil/bin/net_4_0_Release/Mono.Cecil.dll:
+	$(MSBUILD) external/Mono.Cecil/Mono.Cecil.sln /p:Configuration=net_4_5_Release
+
+clean:
```
jonpryor: With a top-level `bin`, you would add:

    -$(RM) -Rf bin/$(CONFIGURATION)

Makefile
```diff
@@ -0,0 +1,31 @@
...
+	$(MSBUILD) external/SharpZipLib/ICSharpCode.SharpZipLib.NET45/ICSharpCode.SharpZipLib.csproj /t:clean
+	$(MSBUILD) external/Mono.Cecil/Mono.Cecil.sln /t:clean
+
+check: check-monodoc check-mdoc
+
+check-mdoc:
+	cd mdoc; $(MAKE) check
+
+check-monodoc:
+	cd monodoc; $(MAKE) check
```
jonpryor: There should be a trailing blank line in the file.

mdoc/Makefile
```diff
@@ -1,6 +1,7 @@
-thisdir = tools/mdoc
-SUBDIRS = 
-include ../../build/rules.make
+PROGRAM = bin/Release/mdoc.exe
```
jonpryor: With a top level `bin`, this becomes `../bin/$(CONFIGURATION)/mdoc.exe`

mdoc/Mono.Documentation/exceptions.cs
```diff
@@ -75,7 +75,7 @@ public ExceptionLookup (ExceptionLocations locations)
 				if (member == null)
 					throw new ArgumentNullException ("member");
 
-				var memberDef = member.Resolve ();
+				var memberDef = member.Resolve();
```
jonpryor: 1. Whitespace-only changes should be avoided.
2. Mono conventions is space before `(`, so this change is wrong, anyway.

mdoc/mdoc.csproj
```diff
@@ -37,21 +37,21 @@
     <Reference Include="System.Xml.Linq" />
     <Reference Include="ICSharpCode.SharpZipLib" />
     <Reference Include="System.Xml" />
-    <Reference Include="monodoc">
-      <HintPath>..\..\class\lib\net_4_x\monodoc.dll</HintPath>
+    <Reference Include="Mono.Cecil.Rocks">
+      <HintPath>..\external\Mono.Cecil\bin\net_4_0_Release\Mono.Cecil.Rocks.dll</HintPath>
```
jonpryor: With a top level `bin`, `%(Reference.HintPath)` needs to change accordingly:

    <HintPath>..\bin\$(Configuration)\Mono.Cecil.Rocks.dll</HintPath>

monodoc/Test/Monodoc/HelpSourceTests.cs
```diff
@@ -67,7 +67,8 @@ public bool Generate (HelpSource hs, string id, Dictionary<string, string> conte
 		[Test]
 		public void ReachabilityTest ()
 		{
-			var rootTree = RootTree.LoadTree (Path.GetFullPath (BaseDir), false);
+			string path = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(this.GetType().Assembly.Location), BaseDir));
```
jonpryor: 1. Why avoid `var`?
2. Convention is space before `(`:

        var path = Path.GetFullPath (Path.Combine (Path.GetDirectoryName (this.GetType ().Assembly.Location), BaseDir));

monodoc/Test/Monodoc/HelpSourceTests.cs
```diff
@@ -108,7 +109,8 @@ IEnumerable<Node> GetLeaves (Node node)
 		[Test]
 		public void ReachabilityWithShortGenericNotationTest ()
 		{
-			var rootTree = RootTree.LoadTree (Path.GetFullPath (BaseDir), false);
+			string path = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(this.GetType().Assembly.Location), BaseDir));
```
jonpryor: Same issue as above, and given that you now have duplicated path lookup code, perhaps a `static` helper method would be better?

    static RootTree LoadTree ()
    {
        var path = Path.GetFullPath (
                Path.Combine (
                    Path.GetDirectoryName (this.GetType ().Assembly.Location),
                    BaseDir));
        return RootTree.LoadTree (path, false);
    }

monodoc/Test/Monodoc/RootTreeTest.cs
```diff
@@ -13,15 +13,16 @@ namespace MonoTests.Monodoc
...
+		const string BaseDir = "../../monodoc_test/";
 
 		RootTree root;
 		HtmlGenerator generator;
 
 		[SetUp]
 		public void Setup ()
 		{
-			root = RootTree.LoadTree (BaseDir, includeExternal: false);
+			string path = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(this.GetType().Assembly.Location), BaseDir));
```
jonpryor: ...seriously, how many times do we have this expression running around?

monodoc/Test/Test.cs
```diff
@@ -0,0 +1,13 @@
+﻿using NUnit.Framework;
+using System;
+namespace Monodoc.Test
+{
+	[TestFixture()]
+	public class Test
```
jonpryor: This file can be removed.

joelmartinez:
@jonpryor thanks for the review. I've addressed the points you brought up ... aside from a squash of commits, any other feedback?

Makefile
```diff
@@ -0,0 +1,41 @@
...
+BIN = bin/$(CONFIGURATION)
+MDOC = $(BIN)/mdoc.exe
+
+all: build
+
+build: $(MDOC)
+
+$(MDOC): $(BIN)/ICSharpCode.SharpZipLib.dll $(BIN)/Mono.Cecil.dll
+	$(MSBUILD) apidoctools.sln /p:Configuration=$(CONFIGURATION)
+	cp mdoc/$(BIN)/mdoc.* $(BIN)/
```
jonpryor: I think the `.csproj` should build directly into `bin/$(Configuration)`, not have a separate build step.

For the `external/*` directories, this isn't really an option -- you don't want to have to modify them -- but for `.csproj` files in *this* repo, it's straightforward to have them do:

    <OutputPath>..\bin\Debug</OutputPath>

This would avoid the need for an "extra" `cp` in the Makefile, and help keep things consistent.

*Similar* logic *can* apply to the contents of `external`, in that you might be able to override the `$(OutputPath)` property:

    $(MSBUILD) external/SharpZipLib/ICSharpCode.SharpZipLib.NET45/ICSharpCode.SharpZipLib.csproj \
        /p:OutputPath=..\..\..\bin\$(CONFIGURATION) \
        /p:Configuration=Release

However, in my experience that *occasionally* goes "wrong," so you'd need to test+verify first.

Makefile
```diff
@@ -0,0 +1,41 @@
...
+	$(MSBUILD) external/SharpZipLib/ICSharpCode.SharpZipLib.NET45/ICSharpCode.SharpZipLib.csproj /p:Configuration=Release
+	cp external/SharpZipLib/bin/Release/ICSharpCode.SharpZipLib.* $(BIN)/
+
+$(BIN)/Mono.Cecil.dll:
+	$(MSBUILD) external/Mono.Cecil/Mono.Cecil.sln /p:Configuration=net_4_5_Release
+	cp external/Mono.Cecil/bin/net_4_0_Release/Mono.Cecil*.dll $(BIN)/
+
+clean:
+	rm -rf bin/$(CONFIGURATION)
+	rm -rf packages
```
jonpryor: Nuking `packages` should be part of a `distclean` or `really-clean` target. The contents of `packages` *shouldn't* change; all deleting this will do is possibly require that you re-download the package contents.

Makefile
```diff
@@ -0,0 +1,41 @@
...
+$(BIN)/ICSharpCode.SharpZipLib.dll:
+	$(MSBUILD) external/SharpZipLib/ICSharpCode.SharpZipLib.NET45/ICSharpCode.SharpZipLib.csproj /p:Configuration=Release
+	cp external/SharpZipLib/bin/Release/ICSharpCode.SharpZipLib.* $(BIN)/
+
+$(BIN)/Mono.Cecil.dll:
+	$(MSBUILD) external/Mono.Cecil/Mono.Cecil.sln /p:Configuration=net_4_5_Release
+	cp external/Mono.Cecil/bin/net_4_0_Release/Mono.Cecil*.dll $(BIN)/
+
+clean:
+	rm -rf bin/$(CONFIGURATION)
```
jonpryor: Nuking `bin/$(CONFIGURATION)` should be done *after* running the `Clean` targets.

joelmartinez:
Thanks for all your help on this PR, @jonpryor! I'm pretty excited at having everything configured in a way that I can both edit/debug/test in VS *and* via makefile. Looking forward to getting this re-integrated back into mono in the near future :)

Issue #13, opened 2017-01-18
[mdoc] Support different platform variants and versions


dend:
`mdoc` does a good job detecting assembly versions, but those are not always representative of the platform variant and the platform version itself. So, say, .NET Framework 4.0 and 4.5 will both have `mscorlib.dll` at version 4.0 - we need to be able to specify the variant and version during `mdoc` run.

As an example of the proposal:

```
mdoc update *.dll -o TestFolder --platform netfx --platformversion 4.0
```

jonpryor:
Variations on this have been a longstanding "problem"/"desired feature", the primary problem being one of representation. For example, "Desktop .NET", Xamarin.Android, Xamarin.iOS, and Xamarin.Mac all support `System.String.Length`, and three of those have the same assembly versions (v2.0.5.0)! It could be extremely handy to have a single set of "core documentation" shared by *everything*, and the documentation saying "This member is supported in platforms: .NET, Android, iOS, Mac".

Unfortunately, the current `<AssemblyInfo/>`/`<AssemblyVersion/>` blob doesn't really support that at all. Some alternate construct is required.

Additionally, I'm not entirely certain that we want to save that info as per-member data, as it would require that all docs be updated whenever a new platform is introduced.

dend:
@jonpryor this ask is specifically related to an automated process that will generate documentation from pre-defined platform drops. I agree that this goes slightly beyond `<AssemplyInfo>` and potentially needs another node along the lines of:

```
<PlatformInfo>
     <PlatformVersion Variant="netfx" Version="4.0" />
     <PlatformVersion Variant="netfx" Version="4.5" />
     <PlatformVersion Variant="netfx" Version="3.5" />
     <PlatformVersion Variant="netfx" Version="2.0" />
     <PlatformVersion Variant="xamarinbcl" Version="1.0" />
</PlatformInfo>
```

joelmartinez:
Yes indeed ... the current proposal to address this need is as follows:

-----

The idea is to be able to have a single source of information for all .NET APIs, regardless of the subset that is exposed.

Consider `System.Type` which is part of mscorlib.dll since 1.0, but has evolved by adding more members.   Or consider new data types that have been added in new versions of the frameworks.

What we want is to be able to generate an XML document for each type that annotates in which version the type and members are present.

[...]

Issue #16, opened 2017-01-24
[mdoc] Handle different inheritance hierarchy


dend:
`System.IO.Stream` has different inheritance hierarchy in .NET Fx and Core. See the definition here:
https://github.com/dotnet/coreclr/blob/release/1.1.0/src/mscorlib/src/System/IO/Stream.cs#L34

There is a conditional directive to control whether the `Stream` class derives from `MarshalByRefObject` or not. So in .NET Fx, it’s derived from `MarshalByRefObject` but in .NET core it’s not. 

This should also be handled by `mdoc`.

joelmartinez:
Hmm, so this complicates matters a bit. 

The first question to tackle here, is that we currently have a single file, based on the type name, to represent a type ... unlike at the `<Member>` level, where we can simply make another node. So first let's define what do we *want* to happen here:

1. Do we want to have _two files_ in this case?
    - If so, what do we use for a file naming convention in cases where there is more than one "source of truth" for a given type: `System.IO/Stream.cs` and `System.IO/MarshalByRefObject-Stream.cs`
    - The problem with the approach above is that all downstream clients of this file system will now have to be aware of this new file convention.
    - Also, every other member that is otherwise unaffected will have to have its documentation duplicated in both files.
1. Do we want to have some sort of "framework qualifier" on any given XML node that can specify that something exists only in a given framework?
1. Or, do we want to introduce a new "node type" that could act as a sort of mini-conditional statement ... something like  
    ```xml
    <ConditionalValue Framework="netcore">
      <True>System.MarshalByRefObject</True>
      <False>System.Object</False>
    </ConditionalValue>
    ```


I'm _tentatively leaning_ towards the second solution, where we have something like an attribute called `FrameworkAlternate`, which would mean that the value without the attribute is the one to use by default, and otherwise you should use the one that matches the framework that is being displayed at the moment:

```xml
  <Base>
    <BaseTypeName>System.Object</BaseTypeName>
    <BaseTypeName FrameworkAlternate="netcore">System.MarshalByRefObject</BaseTypeName>
  </Base>
  <Interfaces>
    <Interface>
      <InterfaceName>System.IDisposable</InterfaceName>
    </Interface>
  </Interfaces>
```
In a similar vein, we could introduce a `FrameworkExclude` attribute for situations like attributes, where an attribute is everywhere but a given framework ... as is the case with the `Serializable` attribute, which doesn't exist in `netcore` yet.
```xml
  <Attributes>
[...]

Issue #9, opened 2016-12-11
[mdoc] Microsoft.VisualBasic.Compatibility.Data.dll breaks the tool


dend:
When running `mdoc` on `Microsoft.VisualBasic.Compatibility.Data.dll` (in `C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0`), the tool errors out with:

```
mdoc: Mono.Cecil.AssemblyResolutionException: Failed to resolve assembly: 'MSDATASRC, Version=7.0.3300.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
  at Mono.Cecil.BaseAssemblyResolver.Resolve (Mono.Cecil.AssemblyNameReference name, Mono.Cecil.ReaderParameters parameters) [0x00121] in <364321ccd6674266bec9cd4fab3f2d34>:0
  at Mono.Cecil.BaseAssemblyResolver.Resolve (Mono.Cecil.AssemblyNameReference name) [0x00007] in <364321ccd6674266bec9cd4fab3f2d34>:0
  at Mono.Cecil.DefaultAssemblyResolver.Resolve (Mono.Cecil.AssemblyNameReference name) [0x0002b] in <364321ccd6674266bec9cd4fab3f2d34>:0
  at Mono.Cecil.MetadataResolver.Resolve (Mono.Cecil.TypeReference type) [0x00040] in <364321ccd6674266bec9cd4fab3f2d34>:0
  at Mono.Cecil.ModuleDefinition.Resolve (Mono.Cecil.TypeReference type) [0x00006] in <364321ccd6674266bec9cd4fab3f2d34>:0
  at Mono.Cecil.TypeReference.Resolve () [0x00013] in <364321ccd6674266bec9cd4fab3f2d34>:0
  at Mono.Documentation.DocUtils.GetInheritedInterfaces (Mono.Cecil.TypeDefinition type) [0x0007d] in <ffc07159c23e4de8b7c8063bdf9582a3>:0
  at Mono.Documentation.DocUtils.GetUserImplementedInterfaces (Mono.Cecil.TypeDefinition type) [0x00000] in <ffc07159c23e4de8b7c8063bdf9582a3>:0
  at Mono.Documentation.CSharpFullMemberFormatter.GetTypeDeclaration (Mono.Cecil.TypeDefinition type) [0x001a9] in <ffc07159c23e4de8b7c8063bdf9582a3>:0
[...]

PR #18, opened 2017-02-02
mdoc: Support documenting multiple frameworks

joelmartinez:
Per: https://github.com/mono/api-doc-tools/issues/13

- if `-frameworks` is present, mdoc creates a folder called `Frameworks`
- for each _frameworkName_ folder present in `path/to/frameworks`, we create an file; _frameworkName.xml_
- That XML file would contain every type and member that is available in that framework
  - This is similar in concept to the [`index.xml`](https://github.com/mono/mono/blob/master/mcs/class/System/Documentation/en/index.xml) file, except that the purpose would be to act as a lookup for availability.
  - The list of available frameworks can simply be an enumeration of the files in the `Frameworks` folder.

The file would look as such:

```xml
<Framework Name="corefx" Version="1.0">
  <Type Name="System.Type">
    <Member Signature="public override Type ReflectedType { get; }" Language="C#" />
  </Type>
</Framework>
```


.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "external/SharpZipLib"]
```
jonpryor: This work is looking duplicative of the other PR... :-/

joelmartinez: oh, yes ... sorry, I just started this branch based off the other branch so I can start working on this feature. Once it's merged into master, these changes will no longer show up ... and if I change it based on your review/feedback, I'd simply rebase and `--force` push this branch.

mdoc/Mono.Documentation/Frameworks/FrameworkIndex.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.IO;
+using System.Linq;
+using System.Xml;
+using System.Xml.Linq;
+
+using Mono.Cecil;
+
+namespace Mono.Documentation
+{
+	public class FrameworkTypeEntry : IComparable<FrameworkTypeEntry>
```
jonpryor: Does this need to be public?

mdoc/Mono.Documentation/Frameworks/FrameworkIndex.cs
```diff
@@ -0,0 +1,163 @@
...
+
+		public string Name { get; set; }
+
+		public ISet<string> Members {
+			get {
+				return this.members;
+			}
+		}
+
+		public virtual void ProcessMember (string sig) {
```
jonpryor: Why the divergence from the normal Mono coding style? `{` on a new line.

mdoc/Mono.Documentation/Frameworks/FrameworkIndex.cs
```diff
@@ -0,0 +1,163 @@
...
+
+			var entry = frameworks.FirstOrDefault (f => f.Name.Equals (shortPath));
+			if (entry == null) {
+				entry = new FrameworkEntry { Name = shortPath };
+				frameworks.Add (entry);
+			}
+			return entry;
+		}
+
+		public void Write (string path) {
```
jonpryor: More descriptive name please. Write *what*? `WriteFrameworkIndex()`?

mdoc/Mono.Documentation/Frameworks/FrameworkIndex.cs
jonpryor: I'm also not at all sure why this needs to go into a `Frameworks` subdirectory...

mdoc/Test/en.expected-frameworks/FrameworksIndex/One.xml
```diff
@@ -0,0 +1,16 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Framework Name="One">
+  <Type Name="MyFramework.MyNamespace.MyClass">
+    <Member Sig="public double OnlyInClassic { get; set; }" />
```
jonpryor: Naming quibble: `Sig` should be explicit about what it is, particularly since `mdoc` supports both C# and IL-based signature generators. I'd suggest `CSharpSignature`.

jonpryor: I also half wonder if we should use IL for the signature, not C#, because C# can/will change as new features are added to the language.

For example, C# 7 tuples:

    public (int start, int end) GetExtents () {...}

There's two -- compatible! -- ways to render this for doc purposes:

    // mdoc doesn't know wtf tuples are yet...
    public ValueTuple<int, int> GetExtents();

    // mdoc *does* know wtf tuples are
    public (int start, int end) GetExtents();

So long as everything is updated with the same tooling at the same time, and is thus internally consistent, things will work -- updating the C# signature generator to properly support tuples won't "break" anything.

*As soon as* that's not the case -- e.g. documentation comes from multiple different sources and is "integrated" at runtime, which the MonoDoc GUI tries to do on macOS -- then you may encounter problems.

Offhand, I'm not sure if this "frameworks" support is something that could plausibly support "merge at runtime from various disparate sources" behavior.

If it *can*, that means we need a *stable*, *unchanging* signature language.

*C# is not that language*. IL *is*.

joelmartinez:
@jonpryor thanks for the code review, I'll adjust. In the meantime, @TianqiZhang has brought up an interesting question. How should the `-lib\-L` lookup directory functionality work, in the context of this feature?

After some thought, what do you think of having an `AssemblyResolver` for every framework ... and then a simple convention-based config file in the root folder of every framework folder; say, `FrameworksIndex/<framework>/mdoc.xml` ... that config file could look something like this:

```xml
<mdoc>
   <lookupDirectory>/path/to/dependencies</lookupDirectory>
</mdoc>
```

Thoughts?

/cc @dend

jonpryor:
@joelmartinez: I didn't see @TianqiZhang's question. I'm guessing it's Issue #21?

If so, a question: how often are we going to hit "incompatible" dependencies in which it actually *matters*, documentation-wise, that we resolve the "right" assemblies? Have a specific example in mind?

> After some thought, what do you think of having an `AssemblyResolver` for every framework

This would work, but starts fundamentally changing the mental model: we're not *solely* using command-line arguments -- and files via `@include.file` syntax -- we're using a config file of some form. Where is this file located? How does it interact with the command-line args?

For simplicity, I would far prefer that everything remain command-line based. If that can't be supported, we should know *precisely* **why** it can't be supported.

TianqiZhang:
@jonpryor Actually it about issue #9. To workaround #9 we need to use -L to specify a dependency folder. However in #13 -frameworks mode, mdoc will be running against multiple platform/versions. How does -L work in this case?

jonpryor:
> How does -L work in this case?

How does it not work?

To be fair, it'll be *weird*, in that you'd need to `-L PATH` *all* dependencies for *all* frameworks directories. Then the only problem is that the first assembly found in those directories is the one that's used.

Is that a problem?

Yes!

> If [command-line only] can't be supported, we should know precisely why it can't be supported.

I must have needed sleep then. *Now* I can imagine how "just use `-L PATH`!" will fail, and it's already mentioned: Issue #16.

Sometimes base classes vary based on the framework, and those differences could be in `mscorlib.dll` (e.g. `System.IO.Stream`). In those cases, which `mscorlib.dll` is resolved matters greatly as to the resulting API description.

jonpryor:
@joelmartinez: How's this for an alternate take: instead of an `mdoc update -frameworks DIRECTORY`, we instead require `mdoc update -frameworks CONFIG_FILE`?

`mdoc update -L PATH` should be "global". I'm not sure how useful that is, but it'll help with the mental model: parameters on the command line apply to *all* frameworks.

The `-frameworks FILE` config file would need to pull multiple duties (which is why it's a config file):

* The directory the framework files are in
* Assembly search paths
* ...anything else?

As a starting point for discussion purposes -- and not thought through!

```xml
<frameworks>
  <framework name="netcore" source="frameworks/netcore">
    <assemblySearchPath>lib/netcore</assemblySearchPath>
  </framework>
</frameworks>
```

This seems somewhat overblown for now, in that we "only" need to tie together two values...except `//assemblySearchPath` could be *multiple* elements, so it's *not* "just two values".

dend:
@jonpryor @joelmartinez I like the idea of using a config file for complex scenarios. This also gives us some leeway in the future to not clutter the command line and just use the configuration. Given that in cases when one wants to reflect from several frameworks at once, the user is surely accounting for the fact that there might be some delta in dependencies, even with a common core, so separating those would be good.

dend:
Looking at the file, would be great to also get the namespaces in the breakdown of framework contents. Something like this (as mentioned by @joelmartinez):

```
<Framework Name="corefx" Version="1.0"> 
 <Namespace Name="System">
  <Type Name="System.Type">
    <Member Signature="public override Type ReflectedType { get; }" Language="C#" />
  </Type>
 </Namespace>
</Framework>
```

mdoc/Test/en.expected-frameworks/FrameworksIndex/One.xml
joelmartinez: Since you mentioned that you'd be fine with [adding the "DocId" in v5.0 for everyone](https://github.com/mono/api-doc-tools/issues/31#issuecomment-282885080), and per @dend's request ... I'm going to use this new "DocId" as the identifier for the FrameworksIndex files. 

Mentioning here in this thread to give a chance for thoughts if there are any :)

Issue #8, opened 2016-12-11
[mdoc] System.EnterpriseServices.* are causing breaks


dend:
If you try running `mdoc` on a range of assemblies in the .NET Framework (by default, located in, say `C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0`), assemblies in the `System.EnterpriseServices.*` range are causing tool errors (e.g. `System.EnterpiseServices.Wrapper.dll`):

```
mdoc: System.ArgumentException: Value does not fall within the expected range.
  at Mono.Cecil.AssemblyDefinition.ReadAssembly (Mono.Cecil.ModuleDefinition module) [0x0000d] in <364321ccd6674266bec9cd4fab3f2d34>:0
  at Mono.Cecil.AssemblyDefinition.ReadAssembly (System.String fileName, Mono.Cecil.ReaderParameters parameters) [0x00007] in <364321ccd6674266bec9cd4fab3f2d34>:0
  at Mono.Documentation.MDocUpdater.LoadAssembly (System.String name) [0x00015] in <ffc07159c23e4de8b7c8063bdf9582a3>:0
  at Mono.Documentation.MDocUpdater+<Run>c__AnonStorey2.<>m__E (System.String a) [0x00000] in <ffc07159c23e4de8b7c8063bdf9582a3>:0
  at System.Linq.Enumerable+WhereSelectListIterator`2[TSource,TResult].MoveNext () [0x00064] in <63992662b765477a898ef49cdcc99ee2>:0
  at System.Collections.Generic.List`1[T]..ctor (System.Collections.Generic.IEnumerable`1[T] collection) [0x0008b] in <8f2c484307284b51944a1a13a14c0266>:0
  at System.Linq.Enumerable.ToList[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00011] in <63992662b765477a898ef49cdcc99ee2>:0
  at Mono.Documentation.MDocUpdater.Run (System.Collections.Generic.IEnumerable`1[T] args) [0x003c0] in <ffc07159c23e4de8b7c8063bdf9582a3>:0
  at Mono.Documentation.MDoc.Run (System.String[] args) [0x001f6] in <ffc07159c23e4de8b7c8063bdf9582a3>:0
  at Mono.Documentation.MDoc.Main (System.String[] args) [0x00006] in <ffc07159c23e4de8b7c8063bdf9582a3>:0
```

joelmartinez:
[...]

PR #23, opened 2017-02-15
mdoc: Failing to load an assembly is no longer a fatal error.

### LINKED ISSUE #7: [mdoc] Error when running on a folder with some assemblies being non-managed
For example, if you run:

```
mdoc update *.dll -o folder_name
```
And within the folder (current context) there are DLLs that are native, the tool will fail with:

```
Format of the executable (.exe) or library (.dll) is invalid
```

joelmartinez:
In some cases, for example when running in "frameworks" mode (ie. https://github.com/mono/api-doc-tools/pull/18), you might have non-managed .dlls in the directory. This change allows mdoc to simply write the fact that it couldn't load the assembly, and continue doing work.
Resolves #7
Provices a temporary workaround for #8, #9

### DIFF
--- a/mdoc/Mono.Documentation/monodocer.cs
+++ b/mdoc/Mono.Documentation/monodocer.cs
@@ -320,8 +320,11 @@ class MDocUpdater : MDocCommand
 		if (srcPath == null)
 			throw new InvalidOperationException("The --out option is required.");
 		
-		this.assemblies = assemblies.Select (a => LoadAssembly (a)).ToList ();
-
+		this.assemblies = assemblies.Select (a => LoadAssembly (a)).Where(a => a != null).ToList ();
+		
+		if (assemblies.Count == 0)
+			Error ("No valid assemblies specified.");
+			
 		// Store types that have been forwarded to avoid duplicate generation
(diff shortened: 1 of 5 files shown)


mdoc/Mono.Documentation/monodocer.cs
```diff
@@ -408,10 +411,10 @@ private AssemblyDefinition LoadAssembly (string name)
 		AssemblyDefinition assembly = null;
 		try {
 			assembly = AssemblyDefinition.ReadAssembly (name, new ReaderParameters { AssemblyResolver = assemblyResolver });
-		} catch (System.IO.FileNotFoundException) { }
-
-		if (assembly == null)
-			throw new InvalidOperationException("Assembly " + name + " not found.");
+		} 
+		catch (Exception ex) {
+			Warning ($"Error loading assembly '{name}': {ex.Message}");
```
jonpryor: If it's a *warning*, then you shouldn't say "Error"; that can make things confusing. I would suggest:

    Warning ($"Unable to load assembly `{name}`: {ex.Message}");

PR #34, opened 2017-03-04
Support for differing inheritance chains in multiple frameworks

joelmartinez:
When a class has a different inheritance chain in two different frameworks, the changes in this PR will support that by adding a `FrameworkAlternate` parameter to the `BaseTypeName` node for the alternate parameter.

```xml
  <Base>
    <BaseTypeName>MyNamespace.MyBaseClassOne</BaseTypeName>
    <BaseTypeName FrameworkAlternate="Two">MyNamespace.MyBaseClassTwo</BaseTypeName>
  </Base>
```

Per Issue #16

### DIFF
--- a/mdoc/.gitignore
+++ b/mdoc/.gitignore
@@ -2,6 +2,6 @@
 /Test/html.*/
 /Test/DocTest.*
 /Test/*.dll*
-/Test/FrameworkTestData/
+/Test/FrameworkTestData*
 /.v2.txt
 /.v0.txt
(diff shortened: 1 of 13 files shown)


joelmartinez:
@dend so I've implemented this ... however it comes with a bit of a caveat that I just wanted to highlight; it's dependent on starting conditions and order of operations. Here's what I mean:

- The _first_ framework in the `frameworks.xml` that is processed will serve as the "canonical BaseTypeName". So in the example provided in the original GH issue, you will want to process the .NET framework first in the config file, since .NET core is the one that is "alternate".
- If you're updating an existing set of XMLs, the old behavior was to replace the `BaseTypeName` with the last framework processed. 

So basically, just keep an eye on this to make sure that the resulting XML is what you expect ... once you run it the first time, subsequent updates will have consistent results.

Issue #20, opened 2017-02-03
EcmaUrlParser Parser Grammar


joelmartinez:
The [`EcmaUrlParser`] is built using Jay and this [grammar](monodoc/Monodoc.Ecma/EcmaUrlParser.jay). For the time being, the [prebuilt version of `EcmaUrlParser.cs`](monodoc/Monodoc.Ecma/prebuilt/EcmaUrlParser.cs) is included in this repository, but we need to move to building from the grammar in order to allow future edits to the format, in support of https://github.com/mono/api-doc-tools/issues/17

joelmartinez:
@jonpryor Can you give me a little bit of guidance on how we might consume this dependency? Are we using this [C# port of Jay](https://github.com/mono/mono/tree/0bcbe39b148bb498742fc68416f8293ccd350fb6/mcs/jay), in the mono repo?

If so, what would be the easiest way to move ahead, copy those files into this repository? Or it distributed with mono?

jonpryor:
I think this might be a sign that we should consider using a *different* grammar generator. Is there something available via nuget?

joelmartinez:
Will inspect the list later, but it looks like there's a few: http://nugetmusthaves.com/Tag/grammar

joelmartinez:
@dend I've removed this from the [v5.0 Preview], because #17 has also been removed from that milestone in lieu of #35. This work item will only need to be implemented for the full resolution of #17 :)

Issue #17, opened 2017-02-01
[mdoc] Group overloads and allow information related to all of them


dend:
One of the scenarios that we are encountering is the need to associate certain information with all overloads of a member. For example:

https://msdn.microsoft.com/en-us/library/system.string.compare(v=vs.110).aspx 

Notice that there is a description associated with the set of overloads. This is something that is not possible today in ECMAXML as it does not allow descriptions for member sets, or even the grouping of members.

The ask is to allow member grouping per-name (e.g. find overloads of a method) and then allow information to be presented within that group.

joelmartinez:
Two potential approaches, based on a chat convo with @dend ... documenting for discussion:

## Multiple `<Member>` Nodes

Would allow for manual grouping of nodes ... you could add a common `remarks` section that would be rendered first for all member nodes. Something like

```xml
<Members>
    <remarks>...</remarks>
    <Member Name="foo">...</Member>
     ...
</Members>
<Members>
    <remarks>...</remarks>
    <Member Name="bar">...</Member>
    ...
</Members>
```

## A New `<MemberGroup>` Node

A node that would represent common information about a set of overloads

```xml
<Members>
    <MemberGroup Name="foo">...</MemberGroup>
    <Member Name="foo">...</Member>
    <Member Name="foo">...</Member>
</Members>
```

jonpryor:
I'm partial to the `<MemberGroup/>` idea.

That said, I'm conflicted. Consider the all-important [`String.Format()` method](https://msdn.microsoft.com/en-us/library/system.string.format(v=vs.110).aspx). Notice that:

1. The documentation for the method overload page is *huge*; the method overload page *is* the documentation, for most intents and purposes.
2. The documentation for an individual method, such as [`String.Format(IFormatProvider, String, Object)`](https://msdn.microsoft.com/en-us/library/dn906224(v=vs.110).aspx), is (a) comparatively *very* short, and (b) has a **See Also** link to the Format Overload documentation.

These things together mean that not only do we need to provide a way to document overloads, but we need a way to *link to* the overload section *itself*.

I'm not at all sure how to do that. A "logical" thing to do would be to use no parenthesis with an `M:` "url", e.g. `<see cref="M:System.String.Format" />`, but this syntax has already been "taken" in `monodoc.dll`, and would result in a link to a 0-parameter `System.String.Format()` method (which doesn't exist).

[...]

PR #36, opened 2017-03-05
mdoc: Added MemberGroup as valid element

### LINKED ISSUE #35: [mdoc] Add `MemberGroup` to XSD
Per #17, the `Members/MemberGroup` node will be used to hold documentation that pertains to an entire set of overloaded methods. This task is to add it to the XSD.

The `MemberGroup` will look like this:

```xml
<Members>
    <MemberGroup Name="foo">...</MemberGroup>
    <Member Name="foo">...</Member>
    <Member Name="foo">...</Member>
</Members>
```

Elements that are valid inside of `MemberGroup` include 

- `summary`
- `remarks`
- `AssemblyInfo`

You can see an example of an overload summary page here: [System.String.Compare on MSDN](https://msdn.microsoft.com/en-us/library/system.string.compare(v=vs.110).aspx)

joelmartinez:
A MemberGroup can be intermixed amongst the other Member nodes.
It will be sorted to be in front of the other similarly named Members.
It must contain a MemberName attribute, and can have: summary, remarks, and AssemblyInfo nodes.

```xml

    <MemberGroup MemberName="SomeMethod">
      <AssemblyInfo>
        <AssemblyVersion>0.0.0.0</AssemblyVersion>
      </AssemblyInfo>
      <Docs>
        <remarks>To be added.</remarks>
        <summary>To be added.</summary>
      </Docs>
    </MemberGroup>
```

Resolves #35

### DIFF
--- a/mdoc/Resources/monodoc-ecma.xsd
+++ b/mdoc/Resources/monodoc-ecma.xsd
@@ -481,11 +481,22 @@ add masterdoc support?
     </xs:complexType>
   </xs:element>
 
+  <xs:element name="MemberGroup">
+    <xs:complexType>
+      <xs:choice minOccurs="0" maxOccurs="unbounded">
+        <xs:element ref="AssemblyInfo" minOccurs="0" maxOccurs="unbounded" />
+        <xs:element ref="Docs" minOccurs="0" maxOccurs="1" />
+      </xs:choice>
+      <xs:attribute ref="MemberName" />
+    </xs:complexType>
+  </xs:element>
+
   <xs:element name="Members">
     <xs:complexType>
-      <xs:sequence>
+      <xs:choice minOccurs="0" maxOccurs="unbounded">
+        <xs:element ref="MemberGroup" minOccurs="0" maxOccurs="unbounded" />
         <xs:element ref="Member" minOccurs="0" maxOccurs="unbounded" />
-      </xs:sequence>
+      </xs:choice>
     </xs:complexType>
   </xs:element>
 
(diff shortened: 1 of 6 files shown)


joelmartinez:
@dend can you verify that the sample XML above will meet all of your requirements? All of the elements therein are optional, so you can have any mix of them.

mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -481,11 +481,23 @@ add masterdoc support?
     </xs:complexType>
   </xs:element>
 
+  <xs:element name="MemberGroup">
+    <xs:complexType>
+      <xs:choice minOccurs="0" maxOccurs="unbounded">
+        <xs:element ref="AssemblyInfo" minOccurs="0" maxOccurs="unbounded" />
+        <xs:element ref="summary" minOccurs="0" maxOccurs="1" />
```
jonpryor: Should `MemberGroup` instead include `Docs`? This would make `MemberGroup` more similar to `Member`, and I *suspct* would simplify the XSLT for HTML rendering -- which I notice that this PR is lacking.

joelmartinez:
@jonpryor good point, I changed it to use `Docs`.

As to the missing rendering, that is purposely omitted. I explained my reasoning [here](https://github.com/mono/api-doc-tools/issues/17#issuecomment-284415477) ... and I will follow up [Milestone 1](https://github.com/mono/api-doc-tools/milestone/1) with another set of releases that includes #17 (ie. rendering of MemberGroup nodes), and #20 (ie. modifying the mdoc url grammar) :)

dend:
Thanks for working on this, @joelmartinez

PR #37, opened 2017-03-06
mdoc: special characters in attributes are now filtered.

### LINKED ISSUE #32: [mdoc][xml] Generated XML contains 0x00 that breaks the XDocument parser
Available in files such [as this](https://apidrop.visualstudio.com/_git/XMLDevDrop?path=%2Ffeb-24-2017%2FSystem.Windows.Forms%2FMaskedTextBox.xml&version=GBmaster&_a=contents).

Getting the following errors when trying to use `XDocument.Load(string fileName)`:

```
Failed to load E:\apidrop\XMLDevDrop\feb-24-2017\System.Windows.Forms\MaskedTextBox.xml: '.', hexadecimal value 0x00, is an invalid character. Line 1168, position 65.
Failed to load E:\apidrop\XMLDevDrop\feb-24-2017\System.Windows.Forms\TextBox.xml: '.', hexadecimal value 0x00, is an invalid character. Line 417, position 65.
```

Both files are generated out of the `System.Windows.Forms.dll` assembly in .NET Framework.

joelmartinez:
Fixes #32, which was an issue with the ascii null character. It caused the resulting XML to be
malformed. So instead of being written as `&#x0;`, it will now be written as `\0`. All the characters being filtered in attributes are:

- \0
- \t
- \n
- \r
- \f
- \b

### DIFF
--- a/mdoc/Mono.Documentation/monodocer.cs
+++ b/mdoc/Mono.Documentation/monodocer.cs
@@ -2511,6 +2511,41 @@ class MDocUpdater : MDocCommand
 		"System.Runtime.CompilerServices.DynamicAttribute",
 	};
 
+	private IEnumerable<char> FilterSpecialChars (string value)
+	{
+		foreach (char c in value) {
+			switch (c) {
+				case '\0':
+					yield return '\\';
+					yield return '0';
+					break;
+				case '\t':
+					yield return '\\';
+					yield return 't';
+					break;
+				case '\n':
+					yield return '\\';
+					yield return 'n';
+					break;
+				case '\r':
+					yield return '\\';
+					yield return 'r';
+					break;
+				case '\f':
+					yield return '\\';
+					yield return 'f';
+					break;
+				case '\b':
+					yield return '\\';
+					yield return 'b';
+					break;
+				default:
+					yield return c;
+					break;
+			}
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


mdoc/Mono.Documentation/monodocer.cs
```diff
@@ -2511,6 +2511,41 @@ static bool UpdateAssemblyVersionForAssemblyInfo (XmlElement e, XmlElement root,
 		"System.Runtime.CompilerServices.DynamicAttribute",
 	};
 
+	private IEnumerable<char> FilterSpecialChars (string value)
```
jonpryor: Why'd you go for an `IEnumerable<char>` return type and not `string`? Then you could do e.g. `return @"\0";`, and the calling code wouldn't require a `new string (...)` intermediary.

joelmartinez: Because the incoming string isn't simply `"\0"`, it's the whole attribute value that gets written to the XML (`"System.ComponentModel.DefaultValue('\0')"`). So I'm just enumerating through each char in the string, and when it encounters one of the special chars, I return two chars in its place. `yield return` is a convenient way of doing that :)

jonpryor: Ah. OK.

jonpryor:
LGTM.

PR #41, opened 2017-03-14
mdoc: Resolves issue in fx mode with overridden members.

### LINKED ISSUE #39: FrameworksIndex files contain members of base types in inheriting types
In the `FrameworksIndex` files, classes that inherit from base types will include the members of base types - this should not be the case, as it should be a representation of just the members belonging to the type and nothing else.

joelmartinez:
When a type overrides a member in one framework, but does not in another framework
processed after the first, the Member node was being removed, even though the entry
remained in teh first framework index file. This Resolves #39

### DIFF
--- a/mdoc/Consts.cs
+++ b/mdoc/Consts.cs
@@ -4,6 +4,6 @@ namespace Mono.Documentation
 	public static class Consts
 	{
 		// this is only a placeholder
-		public static string MonoVersion = "5.0.0.7";
+		public static string MonoVersion = "5.0.0.8";
 	}
 }
--- a/mdoc/Mono.Documentation/Frameworks/FrameworkTypeEntry.cs
+++ b/mdoc/Mono.Documentation/Frameworks/FrameworkTypeEntry.cs
@@ -10,7 +10,7 @@ namespace Mono.Documentation
 		SortedSet<string> members = new SortedSet<string> ();
 		SortedSet<string> memberscsharpsig = new SortedSet<string> ();
 
-		CSharpFullMemberFormatter formatter = new CSharpFullMemberFormatter ();
+		ILMemberFormatter formatter = new ILMemberFormatter ();
 
 		FrameworkEntry fx;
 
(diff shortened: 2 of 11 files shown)


mdoc/Consts.cs
```diff
@@ -4,6 +4,6 @@ namespace Mono.Documentation
 	public static class Consts
 	{
 		// this is only a placeholder
-		public static string MonoVersion = "5.0.0.7";
+		public static string MonoVersion = "5.0.0.8";
```
jonpryor: This is crying out for some make voodoo. ;-)

mdoc/Mono.Documentation/Frameworks/FrameworkTypeEntry.cs
```diff
@@ -8,6 +8,10 @@ namespace Mono.Documentation
 	class FrameworkTypeEntry : IComparable<FrameworkTypeEntry>
 	{
 		SortedSet<string> members = new SortedSet<string> ();
+		SortedSet<string> memberscsharpsig = new SortedSet<string> ();
+
+		CSharpFullMemberFormatter formatter = new CSharpFullMemberFormatter ();
```
jonpryor: Do we really want to use the `CSharpFullMemberFormatter` here, and not the IL formatter?

joelmartinez: good point ... there wasn't really any rhyme or reason to it, other than the fact that the C# sig is used in several other spots for bookkeeping purposes and I just defaulted to it. But you're right, `ILSig` is probably a better unit to use going forward :)

mdoc/Consts.cs
joelmartinez: I was just following the convention used before (mdoc just used the same constants file that was generally used in mono, so that's why the version was pegged to the mono release. However, I'm happy to take suggestions ... should the version # be a supplied parameter to the makefile and not stored in source control?

jonpryor: I'd suggest "mimicking" xamarin-android and various other repos:

1. Have a build-time generated file which contains the version number.
2. Version number consists of a hand-curated major.minor.build version, and a generated revision value.
3. The revision value is the number of commits since the last change in the version number.

For example, in xamarin-android the [`$(ProductVersion)` MSBuild property](https://github.com/xamarin/xamarin-android/blob/master/Configuration.props#L15) contains the major.minor.build value, and the revision value [is computed](https://github.com/xamarin/xamarin-android/blob/master/build-tools/scripts/BuildEverything.mk#L7-L11) via `git log <version-change-commit>..HEAD --oneline | wc -l`.

This is particularly handy when you have a CI system building all commits, as every build will have a different, unique, version.

Issue #44, opened 2017-03-20
mdoc adds duplicated EII members to xml when update


TianqiZhang:
Use mdoc 5.0.0.9 to run "mdoc update" with the dlls from below repo:
https://apidrop.visualstudio.com/_git/binaries?path=%2Fdotnet&version=GBmaster&_a=contents
against existing xmls from here:
https://github.com/TianqiZhang/ECMA2YamlTestRepo2/tree/withyml/fulldocset/xml
You'll find that many duplicated members were inserted into xml with exactly the same name, signature and DocId. Take fulldocset\xml\Microsoft.VisualBasic\Collection.xml as an example, after mdoc finishes update, you can find two "System.Collections.ICollection.CopyTo" members and a bunch of other duplicated members.

TianqiZhang:
Here's the full list of duplicate members I found after mdoc update:
[log.json.txt](https://github.com/mono/api-doc-tools/files/853854/log.json.txt)
You can see there are two types, op_Implicit and EII.

TianqiZhang:
I can repro with re-generated xml files from:
https://apidrop.visualstudio.com/_git/binaries?path=%2Fdotnet&version=GBmaster&_a=contents

PR #47, opened 2017-03-23
mdoc: Support for VB Explicitly Implemented Interface members

joelmartinez:
Since VB.NET can explicitly implement a member without using the same
naming convention as you might use in C#, it was causing some issues with
mdoc matching xml member nodes to reflected types. In particular, the cases
supported with this patch are:

- interface type and member name concatenated. e.g. `IListAdd`
- same as above, but when the interface doesn’t match the interface where the member is defined. e.g. `ICollection` inherits `IEnumerable`. So `System.Collections.IEnumerable.GetEnumerator` is expressed in some types as `ICollectionGetEnumerator`
- same name as the interface member e.g. `Microsoft.VisualBasic.Compability.VB6.BaseControlArray$System.ComponentModel.ISupportInitialize.BeginInit`

Related #44

### DIFF
--- a/mdoc/Consts.cs
+++ b/mdoc/Consts.cs
@@ -4,6 +4,6 @@ namespace Mono.Documentation
 	public static class Consts
 	{
 		// this is only a placeholder
-		public static string MonoVersion = "5.0.0.9";
+		public static string MonoVersion = "5.0.0.10";
 	}
 }
--- a/mdoc/Consts.cs
+++ b/mdoc/Consts.cs
@@ -3,6 +3,6 @@ namespace Mono.Documentation
 {
 	public static class Consts
 	{
-		public static string MonoVersion = "5.0.0.11";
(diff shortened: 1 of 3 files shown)


joelmartinez:
Added the `work-in-progress` label because there's still one more case that needs to be tested for `op_Implicit` members that are being duplicated. So far I've been unable to reproduce.

joelmartinez:
@dend, @TianqiZhang although I thought that had reproduced the `op_Implicit` issue, in actuality I hadn't (the signatures just looked similar, but they weren't duplicated). So with the [v5.0.0.10](https://github.com/mono/api-doc-tools/releases/tag/preview-5.0.0.10) release, and working with the assemblies mentioned in the json log ... it seems to have resolved all of the issues in my testing.

Please let me know if there are still duplicate nodes after multiple runs with this build.

Issue #53, opened 2017-03-28
In frameworks mode, mdoc should not mark any types as removed if they exist in another framework


TianqiZhang:
When running againse existing xml files, "mdoc update" will rename some types to ".remove" if they don't exist in newer frameworks.

That's not expected because in -fx mode, we have the framework index files to tell which type belongs to which framework, so none of the types should be "removed".

TianqiZhang:
Take `System.Windows.Markup.WhitespaceSignificantCollectionAttribute` for example, it existed in .Net framework 3.0, 3.5 and 4.0. However after mdoc update, its xml file was renamed to .removed.

joelmartinez:
Hmm ... so far in my tests, I haven't been able to reproduce this. I modified a test to run multiple times, with types that exist in different frameworks only.

joelmartinez:
Ok ... I was finally able to reproduce this successfully. Will address now

PR #56, opened 2017-03-28
mdoc: no longer skips documenting forwarded types

joelmartinez:
Previously, mdoc simply skipped documenting forwarded types. With this change, forwarded types will still be documented if they exist in the assemblies being documented ... and they will contain the assembly info of the Assembly where the types are declared.


joelmartinez:
This resolves #53

Issue #71, opened 2017-04-21
monodoc-ecma.xsd does not reflect current usage


lobrien:
The monodoc-ecma.xsd in mdoc/Resources is missing some elements / attributes that are broadly used in our existing (large) monodoc directories. Importantly, it is missing `apistyle`, `attribution`, and`tool`. 

If we are to move towards broader usage of monodoc / mdoc, we need to, at the least, have an accurate set of type definitions.

joelmartinez:
This is resolved in https://github.com/mono/api-doc-tools/pull/72 (when that is merged) ... follow that PR for status

PR #72, opened 2017-04-21
“Facts on the ground” XSD matching ios-api-docs

lobrien:
Relates to https://github.com/xamarin/ios-api-docs/issues/44 & https://github.com/xamarin/DocWriter/issues/11

Closes https://github.com/mono/api-doc-tools/issues/71

With this, it would be possible to add, e.g., `xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation=“https://github.com/xamarin/api-doc-tools/mdoc/Resources/monodoc-ecma.xsd”` to the root `Type` element.


lobrien:
It's worth noting that this is by no means a "perfect" monodoc-xsd. It is definitely bloated and the same exercise of applying it to all the XML in our larger codebases should be done. After we do that, we should start tightening it up.

mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -71,7 +78,6 @@ add masterdoc support?
   <xs:element name="permille" />
   <xs:element name="pi" />
   <xs:element name="pre" />
-  <xs:element name="ReturnType" type="xs:string" />
```
joelmartinez: Just curious ... why is this removing `ReturnType` ?

mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -152,6 +180,7 @@ add masterdoc support?
       <xs:choice minOccurs="0" maxOccurs="unbounded">
         <xs:element ref="block" />
         <xs:element ref="c" />
+        <!-- bryan: should this be allowed? code without example? -->
```
joelmartinez: We can probably remove the "named" comments

mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -162,6 +191,8 @@ add masterdoc support?
         <xs:element ref="sup" />
         <xs:element ref="pi" />
         <xs:element ref="typeparamref" />
+        <!-- bryan -->
```
joelmartinez: another candidate for removal

mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -100,7 +118,8 @@ add masterdoc support?
         <xs:element ref="AssemblyCulture" minOccurs="0" />
         <xs:element ref="Attributes" minOccurs="0" />
       </xs:sequence>
-    </xs:complexType>
+      <xs:attribute ref="apistyle"/>
+     </xs:complexType>
```
jonpryor: This indentation looks wrong.

mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -478,25 +521,15 @@ add masterdoc support?
       </xs:sequence>
       <xs:attribute ref="MemberName" />
       <xs:attribute ref="Deprecated" />
-    </xs:complexType>
-  </xs:element>
-
-  <xs:element name="MemberGroup">
```
jonpryor: I don't think we want to remove `<MemberGroup/>`. See #35, #36.

mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -121,6 +140,15 @@ add masterdoc support?
       </xs:sequence>
     </xs:complexType>
   </xs:element>
+  
+  <!-- License reference -->
+  <xs:element name="attribution">
```
jonpryor: Why move this declaration?

mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -121,6 +140,15 @@ add masterdoc support?
...
+    <xs:complexType>
+    <xs:attribute ref="license"/>
```
jonpryor: This should be indented another "tab stop" (2 spaces).

mdoc/Resources/monodoc-ecma.xsd
joelmartinez: Ahh, yes ... will definitely need to retain this

directhex:
build

PR #74, opened 2017-04-21
mdoc now supports `see href`

joelmartinez:


mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -698,11 +699,16 @@ add masterdoc support?
...
   </xs:element>
 
-  <xs:element name="see">
+ <xs:element name="see">
     <xs:complexType>
-      <xs:attribute ref="cref" />
-      <xs:attribute ref="langword"  />
-      <xs:attribute ref="qualify"  />
+      <xs:simpleContent>
+        <xs:extension base="xs:string">
```
jonpryor: Do we really want to constrain this to just "strings"? I think being able to nest further content would be helpful; not *all* content, certainly, but formatting content such as `<c/>`, `<i/>`, etc., could be reasonable.

PR #90, opened 2017-06-20
Mdoc import support for frameworks mode

joelmartinez:
You can now add a new element to frameworks.xml, `/Frameworks/Framework/import`, which is a path to the import documentation file. An example test case was added in the make target, `check-monodocer-import-fx`, and you can see the sample configuration file here:

```xml
<?xml version="1.0" encoding="utf-8"?>
<Frameworks>
  <Framework Name="one" Source="one">
    <assemblySearchPath>dependencies/one</assemblySearchPath>
    <import>TestEcmaDocs.xml</import>
  </Framework>
  <Framework Name="two" Source="two">
    <assemblySearchPath>dependencies/two</assemblySearchPath>
    <import>TestEcmaDocs2.xml</import>
  </Framework>
</Frameworks>
```

### DIFF
--- a/mdoc/.gitignore
+++ b/mdoc/.gitignore
@@ -3,5 +3,7 @@
 /Test/DocTest.*
 /Test/*.dll*
 /Test/FrameworkTestData*
+/Test/fx-import
+Test/DocTest-DropNS-classic.xml
 /.v2.txt
 /.v0.txt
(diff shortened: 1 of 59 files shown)


joelmartinez:
The latest commit/push/force, adds support for multiple `<import>` elements, which should correspond to the `/doc` output of each assembly in the "framework".

PR #96, opened 2017-07-06
Refactored DocStat, plus added stub for comparereport command

MichaelNorman:
Refactored DocStat, plus added stub for comparereport command. The refactor may touch on others' work in DocStat, so I'm PRing now to get that in, in spite of the indeterminate state of CompareFixCommand.cs.


joelmartinez:
I really like how you added the test XML data simply as files to be copied to the output directory. Going to change my unit tests to reflect the same patter :)

PR #127, opened 2017-10-04
DocStat: Moved changed flag to file loop, out of member loop

MichaelNorman:
I was only catching files where the last member in the file had a change. Moving the changed sigil up to the file level fixed this.


joelmartinez:
whitelabel

joelmartinez:
build

PR #129, opened 2017-10-11
Fix infinite loop in xsl in maketypelink

jdasilva:
I found that under a couple of circumstances that the current xsl would get stuck in an infinite loop.  I was generating html docs for the EntityFramework v6.1.3 nuget package.  The one case I narrowed down occurs on the CurrentValue property in type DbComplexPropertyEntry<TEntity, TComplexProperty> from the base class  DbMemberEntry<TEntity, TProperty> when it attempts to make a link for "TProperty".  It looks like this might have something to do with the generic parameter type name changing from the base class, but I didn't get to the point of definitively understanding the cause.  This change prevents the infinite loop for this and at least one other scenario it hit in the same nuget.  There is probably a more complete fix to be worked out for generating this link, but I felt it was worth it to submit this to prevent the infinite loop.


jdasilva:
Resubmitting on a branch - oops.

Issue #100, opened 2017-07-17
[mdoc] Analytics mode


dend:
When `mdoc` runs, it would be helpful to get some statistics on what was actually produced. Example metrics would include:

* Total types generated from a framework
* Total namespaces generated from a framework

If that information is produced to a CSV with the help of a flag, that would solve the ask.

joelmartinez:
@dend do you want this to be the entire sum of types and namespaces ... or do you want it to be differential, which is to say, how many were added/removed since the last run ?

dend:
@joelmartinez I think it would be interesting to get both. Something along the lines of:

```
framework-1.0
------------
Types: X
Types Added: 0
Types Removed: 0

Namespaces: Y
Namespaces Added: 0
Namespaces Removed: 0

Members: Z
Members Added: 0
Members Removed: 0

framework-2.0
------------
Types: X
Types Added: 0
Types Removed: 0

Namespaces: Y
Namespaces Added: 0
Namespaces Removed: 0

Members: Z
Members Added: 0
Members Removed: 0
```

PR #140, opened 2017-10-23
Mdoc analytics mode

katebialiatka:
Pull request for issue #100 [mdoc] Analytics mode

### DIFF
--- a/mdoc/Mono.Documentation/MDocUpdater.cs
+++ b/mdoc/Mono.Documentation/MDocUpdater.cs
@@ -13,6 +13,7 @@ using System.Xml.XPath;
 using Mono.Cecil;
 using Mono.Documentation.Updater;
 using Mono.Documentation.Updater.Frameworks;
+using Mono.Documentation.Updater.Statistics;
 using Mono.Documentation.Util;
 using Mono.Options;
 
@@ -63,6 +64,8 @@ namespace Mono.Documentation
 
         public static string droppedNamespace = string.Empty;
 
+        private HashSet<string> memberSet;
+
         public static bool HasDroppedNamespace (TypeDefinition forType)
         {
             return HasDroppedNamespace (forType.Module);
@@ -102,11 +105,16 @@ namespace Mono.Documentation
         string FrameworksPath = string.Empty;
         FrameworkIndex frameworks;
         FrameworkIndex frameworksCache;
+        IEnumerable<XDocument> oldFrameworkXmls;
+
+        private StatisticsCollector statisticsCollector = new StatisticsCollector();
 
         static List<string> droppedAssemblies = new List<string> ();
 
         public string PreserveTag { get; set; }
         public bool DisableSearchDirectoryRecurse = false;
+        private bool statisticsEnabled = false;
+        private string statisticsFilePath;
         public static MDocUpdater Instance { get; private set; }
         public static bool SwitchingToMagicTypes { get; private set; }
 
@@ -129,7 +137,10 @@ namespace Mono.Documentation
                 "             for NEW types/members\n" +
                 "If nothing is specified, then only exceptions from the member will " +
                 "be listed.",
-                v => exceptions = ParseExceptionLocations (v) },
+                v =>
+                {
+                    exceptions = ParseExceptionLocations(v);
+                } },
             { "f=",
                 "Specify a {FLAG} to alter behavior.  See later -f* options for available flags.",
                 v => {
@@ -205,6 +216,15 @@ namespace Mono.Documentation
             { "disable-searchdir-recurse",
                 "Default behavior for adding search directories ('-L') is to recurse them and search in all subdirectories. This disables that",
                 v => DisableSearchDirectoryRecurse = true },
+            {
+                "statistics=",
+                "Save statistics to the specified file",
+                v =>
+                {
+                    statisticsEnabled = true;
+                    if (!string.IsNullOrEmpty(v))
+                        statisticsFilePath = v;
+                } },
         };
             var assemblyPaths = Parse (p, args, "update",
                     "[OPTIONS]+ ASSEMBLIES",
@@ -234,6 +254,17 @@ namespace Mono.Documentation
                                                    .ToArray ()
                                   })
                                   .Where (f => Directory.Exists (f.Path));
+
+                oldFrameworkXmls = fxconfig.Root
+                                               .Elements("Framework")
+                                               .Select(f => new
+                                               {
+                                                   Name = f.Attribute("Name").Value,
+                                                   Source = f.Attribute("Source").Value,
+                                                   XmlPath = Path.Combine(srcPath, "FrameworksIndex", f.Attribute("Source").Value + ".xml"),
+                                               })
+                                               .Where(f => File.Exists(f.XmlPath))
+                                               .Select(f => XDocument.Load(f.XmlPath));
 
                 Func<string, string, IEnumerable<string>> getFiles = (string path, string filters) =>
                 {
@@ -326,6 +357,18 @@ namespace Mono.Documentation
 
             if (!string.IsNullOrWhiteSpace (FrameworksPath))
                 frameworks.WriteToDisk (srcPath);
+
+            if (statisticsEnabled)
+            {
+                try
+                {
+                    StatisticsSaver.Save(statisticsCollector, statisticsFilePath);
+                }
+                catch (Exception exception)
+                {
+                    Warning($"Unable to save statistics file: {exception.Message}");
+                }
+            }
 
             Console.WriteLine ("Members Added: {0}, Members Deleted: {1}", additions, deletions);
         }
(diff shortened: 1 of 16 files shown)


joelmartinez:
whitelabel

joelmartinez:
build

mdoc/Mono.Documentation/Updater/Statistics/StatisticsCollector.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Collections.Generic;
+using Mono.Documentation.Updater.Frameworks;
+
+namespace Mono.Documentation.Updater.Statistics
+{
+    public class StatisticsCollector
+    {
+        public Dictionary<string, StatisticsStorage> Storages { get; } =
```
joelmartinez: Please include some triple slash documentation to make it clear, for example, what the `string` key represents

mdoc/Mono.Documentation/Updater/Statistics/StatisticsCollector.cs
```diff
@@ -0,0 +1,21 @@
...
+            new Dictionary<string, StatisticsStorage>();
+
+        public void Increment(string framework, StatisticsItem item, StatisticsMetrics metrics, int delta = 1)
```
joelmartinez: I think a better member name than `Increment`, would be something like `AddMetric`. Mainly because calling `Increment` and passing in a `Removed` just _sounds_ kind of weird.

mdoc/Mono.Documentation/Updater/Statistics/StatisticsStorage.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Mono.Documentation.Updater.Statistics
+{
+    public class StatisticsStorage
```
joelmartinez: Please add some triple-slash documentation with at least a summary description of this class' intended use

mdoc/Mono.Documentation/Updater/Statistics/StatisticsFormatter.cs
```diff
@@ -0,0 +1,31 @@
...
+                var framework = statisticsStoragePair.Key;
+                var statisticsStorage = statisticsStoragePair.Value;
+                result.AppendLine(framework);
+                result.AppendLine("--------");
+                foreach (var statistics in statisticsStorage.Values)
+                {
+                    var staticsItem = statistics.Key;
+                    foreach (var statisticsValuePair in statistics.Value)
+                    {
+                        result.AppendLine(staticsItem + " " + statisticsValuePair.Key + ": " + statisticsValuePair.Value);
```
joelmartinez: Minor thing, but I'd like to see the use of string interpolation here:

```
result.AppendLine($"{staticsItem} {statisticsValuePair.Key}: {statisticsValuePair.Value}");
```

It makes the format of the line easier to read :)

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -1102,6 +1158,17 @@ private void CleanupFiles (string dest, HashSet<string> goodfiles)
                             try { System.IO.File.Delete (newname); } catch (Exception) { Warning ("Unable to delete existing file: {0}", newname); }
                             try { typefile.MoveTo (newname); } catch (Exception) { Warning ("Unable to rename to: {0}", newname); }
                             Console.WriteLine ("Class no longer present; file renamed: " + Path.Combine (nsdir.Name, typefile.Name));
+
```
joelmartinez: Please add some code comments here to explain what's happening.

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -1102,6 +1158,17 @@ private void CleanupFiles (string dest, HashSet<string> goodfiles)
...
+                            var framework = frameworks.Frameworks.FirstOrDefault ();
+                            var frameworkXml = oldFrameworkXmls?.FirstOrDefault
+                                (i => i.XPathSelectElements($"Framework/Namespace/Type[@Name='{typeFullName}']").Any());
+                            var frameworkNameAttribute = frameworkXml?.Root?.Attribute ("Name");
+                            if (frameworkNameAttribute != null)
+                            {
+                                var name = frameworkNameAttribute.Value;
+                                framework = frameworks.Frameworks.FirstOrDefault (i => i.Name == name);
+                            }
+                            statisticsCollector.Increment (framework?.Name ?? "", StatisticsItem.Types, StatisticsMetrics.Removed);
```
joelmartinez: This line has me a bit worried ... there's a lot of uncertainty going on here. What happens if the first parameter is `""`? is an error thrown? Is it covered by a unit test?

This is another case where a code comment would help future maintainers ascertain what's going on.

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -205,6 +213,15 @@ public override void Run (IEnumerable<string> args)
...
                 "Default behavior for adding search directories ('-L') is to recurse them and search in all subdirectories. This disables that",
                 v => DisableSearchDirectoryRecurse = true },
+            {
+                "statistics=",
+                "Save statistics to the specified file",
+                v =>
+                {
+                    statisticsEnabled = true;
+                    if (!string.IsNullOrEmpty(v))
+                        statisticsFilePath = v;
```
joelmartinez: I would like to see output written here, even if it's just a warning, so that the log contains the fact that the filepath is empty. Additionally, should `statisticsEnabled` be set to `false` if the filepath is null or empty (I assume yes)?

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -235,6 +252,17 @@ public override void Run (IEnumerable<string> args)
                                   })
                                   .Where (f => Directory.Exists (f.Path));
 
+                oldFrameworkXmls = fxconfig.Root
```
joelmartinez: Does this feature work at all if you don't use it with a `frameworks` enabled set of documentation? If it does not, please add a check to ensure that frameworks is enabled, and otherwise write a warning to output saying that you had to disable it because it only works with frameworks mode.

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -327,6 +355,12 @@ public override void Run (IEnumerable<string> args)
             if (!string.IsNullOrWhiteSpace (FrameworksPath))
                 frameworks.WriteToDisk (srcPath);
 
+            if (statisticsEnabled)
+            {
+                File.WriteAllText (statisticsFilePath, StatisticsFormatter.Format(statisticsCollector.Storages));
```
joelmartinez: What happens if `statisticsFilePath` doesn't exist? If it automatically creates the file, what happens if the filepath leads to a set of folders that don't exist? What happens if `statisticsFilePath` just points to a folder ... perhaps you could append a default filename?

mdoc/Mono.Documentation/MDocUpdater.cs
mpMelnikov: Joel, thank you again for the detailed comments!
At the moment, as the option is declared "statistics=" passing of the parameter is required. If a user passes just "-statistics", she'll get:
```
mdoc: Missing required value for option '-statistics'.
```
So, even this check `if (!string.IsNullOrEmpty(v))` is excessive now.
We can replace `"statistics="` with `"statisitcs:"`. In this case, this option will not require a parameter. But it seems, that the current version of Mono.Options makes us pass parameters in this format:
`-statistics:some_file.txt`. If I try to pass it as `-statistics some_file.txt`, I get null instead of the passed "some_file.txt".
So, should we make it optional and add some default statistics file name?

joelmartinez: @mpMelnikov it's ok to make it required ... just please do some testing around the behavior of passing in: files that don't exist, files that *do* exist, folders that don't exist (ie. `-statistics some/path/file.txt`), paths that don't exist and cannot exist (ie. a fake drive letter, or path that you don't have permission to write to).

It's ok if the program can't do it, I just want it to fail gracefully, and report a reasonable error message in all cases :)

mdoc/Mono.Documentation/MDocUpdater.cs
joelmartinez: to clarify, what I mean by "no frameworks mode", is just running mdoc without a `frameworks.xml` file, and without the `-fx` parameter. So for example:

```
mdoc update someassembly.dll -o outxml -statistics stats.txt
```

What will happen with the above invocation ... ideally, it would simply work, but if that is not possible (since you're using the framework index files), then at a minimum it should write a warning and make sure that it's disabled so it doesn't fail.

mpMelnikov:
@joelmartinez 
The requested changes are pushed.

mdoc/Mono.Documentation/MDocUpdater.cs
mpMelnikov: This feature works in non-framework mode. In this case, all types are processed with EmptyFrameworkEntry, so we treat it as a usual framework and generate something like this:
```
Framework: Empty
--------
Types Added: 0
Types Removed: 1
Types Total: 3

Namespaces Added: 0
Namespaces Removed: 0
Namespaces Total: 1

Members Added: 1
Members Removed: 6
Members Total: 10
```

joelmartinez:
build

Issue #106, opened 2017-08-08
EII events should not be treated as private


TianqiZhang:
We recently found that mdoc is ignoring Explicit Interface Implementation events, like they were private members. Take [System.Web.UI.WebControls.ListView](https://docs.microsoft.com/en-us/dotnet/api/System.Web.UI.WebControls.ListView?view=netframework-4.7#explicit_interface_implementations_) as an example, according to the [source code](https://referencesource.microsoft.com/#System.Web.Extensions/UI/WebControls/ListView.cs), it has an EII event called `IPageableItemContainer.TotalRowCountAvailable`:
```csharp
[SuppressMessage("Microsoft.Design", "CA1033:InterfaceMethodsShouldBeCallableByChildTypes",
                 Justification = "Unlikely that a derived type would re-implement this event.")]
event EventHandler<PageEventArgs> IPageableItemContainer.TotalRowCountAvailable {
    add {
        Events.AddHandler(EventTotalRowCountAvailable, value);
    }
    remove {
        Events.RemoveHandler(EventTotalRowCountAvailable, value);
    }
}
```

However this member is not in the xml generated by mdoc: [ListView.zip](https://github.com/mono/api-doc-tools/files/1207460/ListView.zip). I doubled checked with ildasm, the dll does contain this member.

dend:
Per internal thread, this is an issue that is causing API losses, therefore a blocking migration bug. We will need to prioritize this high.

joelmartinez:
@TianqiZhang and @dend ... I just wanted to double check that this was still an issue, since I changed the way that mdoc handles EII in release [v5.0.0.24](https://github.com/mono/api-doc-tools/releases/tag/preview-5.0.0.24) on sept 21st. Can we double check?

dend:
I don't see the EII reference in the XML file even with the content generated from the latest version of `mdoc`. @TianqiZhang @mairaw - can you please confirm my assessment?

TianqiZhang:
@joelmartinez @dend I confirm that this is still a issue, we cannot see the expected member described in this issue.

PR #142, opened 2017-10-24
mdoc: Support VB.NET signatures enhancement

### LINKED ISSUE #131: [mdoc] Support VB.NET signatures
Support API entity signatures in Visual Basic.

mpMelnikov:
Added VBFullMemberFormatter and VBMemberFormatter.
Support for -lang CLI parameter (vb.net, docid are available now).
IsSupported virtual methods for TypeReference and MemberReference in MemberFormatter. If they return false, sygnature is not added to XML file.
Closes #131

### DIFF
--- a/mdoc/Test/en.expected-vbnet/MyFramework.MyNamespace/MyClassExtensions.xml
+++ b/mdoc/Test/en.expected-vbnet/MyFramework.MyNamespace/MyClassExtensions.xml
@@ -1,0 +1,41 @@
+<Type Name="MyClassExtensions" FullName="MyFramework.MyNamespace.MyClassExtensions">
+  <TypeSignature Language="C#" Value="public static class MyClassExtensions" />
+  <TypeSignature Language="ILAsm" Value=".class public auto ansi abstract sealed beforefieldinit MyClassExtensions extends System.Object" />
+  <TypeSignature Language="VB.NET" Value="Public Module MyClassExtensions" />
+  <AssemblyInfo>
+    <AssemblyName>DocTest-DropNS-classic</AssemblyName>
+    <AssemblyVersion>0.0.0.0</AssemblyVersion>
+  </AssemblyInfo>
+  <Base>
+    <BaseTypeName>System.Object</BaseTypeName>
+  </Base>
+  <Interfaces />
+  <Docs>
+    <summary>To be added.</summary>
+    <remarks>To be added.</remarks>
+  </Docs>
+  <Members>
+    <Member MemberName="AnExtension">
+      <MemberSignature Language="C#" Value="public static bool AnExtension (this MyFramework.MyNamespace.MyClass value);" />
+      <MemberSignature Language="ILAsm" Value=".method public static hidebysig bool AnExtension(class MyFramework.MyNamespace.MyClass value) cil managed" />
+      <MemberSignature Language="VB.NET" Value="&lt;Extension()&gt;&#xA;Public Function AnExtension (value As MyClass) As Boolean" />
+      <MemberType>Method</MemberType>
+      <AssemblyInfo>
+        <AssemblyName>DocTest-DropNS-classic</AssemblyName>
+        <AssemblyVersion>0.0.0.0</AssemblyVersion>
+      </AssemblyInfo>
+      <ReturnValue>
+        <ReturnType>System.Boolean</ReturnType>
+      </ReturnValue>
+      <Parameters>
+        <Parameter Name="value" Type="MyFramework.MyNamespace.MyClass" RefType="this" />
+      </Parameters>
+      <Docs>
+        <param name="value">To be added.</param>
+        <summary>To be added.</summary>
+        <returns>To be added.</returns>
+        <remarks>To be added.</remarks>
+      </Docs>
+    </Member>
+  </Members>
+</Type>
--- a/mdoc/mdoc.csproj
+++ b/mdoc/mdoc.csproj
@@ -55,6 +55,8 @@
     <Compile Include="Mono.Documentation\Updater\Formatters\StandardFlagsEnumFormatter.cs" />
     <Compile Include="Mono.Documentation\Updater\Formatters\DefaultAttributeValueFormatter.cs" />
     <Compile Include="Mono.Documentation\Updater\Formatters\AttributeValueFormatter.cs" />
+    <Compile Include="Mono.Documentation\Updater\Formatters\VBFullMemberFormatter.cs" />
+    <Compile Include="Mono.Documentation\Updater\Formatters\VBMemberFormatter.cs" />
     <Compile Include="Mono.Documentation\Updater\ResolvedTypeInfo.cs" />
     <Compile Include="Mono.Documentation\Updater\Formatters\FileNameMemberFormatter.cs" />
     <Compile Include="Mono.Documentation\Updater\Formatters\SlashDocMemberFormatter.cs" />
--- a/mdoc/Consts.cs
+++ b/mdoc/Consts.cs
@@ -4,5 +4,9 @@ namespace Mono.Documentation
 	public static class Consts
 	{
 		public static string MonoVersion = "5.2.0.0";
+		public const string DocId = "DocId";
+		public const string VbNet = "VB.NET";
+		public const string DocIdLowCase = "docid";
+		public const string VbNetLowCase = "vb.net";
 	}
 }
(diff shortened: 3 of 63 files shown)


joelmartinez:
whitelist

mdoc/Test/en.expected-vbnet/MyFramework.MyNamespace/MyClassExtensions.xml
```diff
@@ -0,0 +1,41 @@
...
+  <Interfaces />
+  <Docs>
+    <summary>To be added.</summary>
+    <remarks>To be added.</remarks>
+  </Docs>
+  <Members>
+    <Member MemberName="AnExtension">
+      <MemberSignature Language="C#" Value="public static bool AnExtension (this MyFramework.MyNamespace.MyClass value);" />
+      <MemberSignature Language="ILAsm" Value=".method public static hidebysig bool AnExtension(class MyFramework.MyNamespace.MyClass value) cil managed" />
+      <MemberSignature Language="VB.NET" Value="&lt;Extension()&gt;&#xD;&#xA;Public Function AnExtension (value As MyClass) As Boolean" />
```
joelmartinez: Not sure I like the line break and line feed characters. At a minimum, it makes it windows-specific, since other platforms do not use both of these characters to terminate a line.

I'd like for you to check and see if it would be synctactically valid to have the `Extension` on the same line ... ie. `&lt;Extension()&gt; Public Function AnExtension (value As MyClass) As Boolean`

mpMelnikov: You're right, we can have the `<Extension()>` on the same line, I'll delete line break symbols here. But there is a case when we definitely need the new line symbols:
```
Public Class TestClass 
    Inherits TestClass2
    Implements Interface1, Interface2
```
Otherwise, VB.NET compiler shows the error:
```
BC30205	End of statement expected.
```
Should we use some other line ending symbols?

joelmartinez: My initial reaction is to simply use [`Environment.NewLine`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.newline?view=netframework-4.7.1#System_Environment_NewLine), so that at least it respects the conventions of the platform that you're running _mdoc_ on. However ... I suspect there won't be a "good" answer here. 

I'd like to get at least one other opinion on the matter. @jonpryor, @conceptdev, @lobrien, @dend ... Any thoughts on what approach we should take to newline char(s) in VB signatures?

lobrien: My hot-take is “The values of attributes in the ecmadoc must be idempotent / referentially transparent / deterministic. So therefore, Env.NewLine is actually _not_ the correct answer.”

joelmartinez: Another idea that @lobrien had, was to potentially use a character like `¶` (which is encoded in the XML as `&#182;`). This has the benefit of being a single character that downstream tools can easily replace with `Environment.NewLine` before displaying ... but the negative that a simple read of that string won't "just work".

What would be the pros/cons of simply picking one platform's conventions (for example, will `\n` display correctly on a browser on windows, in an html `<pre>` tag)?

joelmartinez: https://codepen.io/anon/pen/gGVmag

In this example, I'm writing `document.write("<pre>this is\nanother line</pre>");`. On mac/chrome this outputs:

image

On windows/IE:

image

I propose that we standardize on `\n` ... thoughts?

joelmartinez: @mpMelnikov can you check and see whether, if we use `\n`, and a user copies and pastes that signature ... will the vb.net compiler throw an error, or will it compile successfully?

dend: Just like @joelmartinez, I suggest we use `\n` as the standard way of handling new lines - this will be universally acceptable for other reference types and follows standard markup conventions for code docs.

mpMelnikov: > can you check and see whether, if we use \n, and a user copies and pastes that signature ... will the vb.net compiler throw an error, or will it compile successfully?

It works fine. I'm writing `document.write("<pre>Public Structure DocValueType\nImplements IProcess</pre>");`, copy result values from Chrome to Visual Studio or Notepad, the line endings become `\n\r`. It seems to be quite a task to save source line endings while copying text.

@joelmartinez If we decide to use `\n`, maybe we can keep new line after `<Extension()>`? That's how it looks like on [docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/procedures/extension-methods) now.

joelmartinez: @mpMelnikov ok, thanks for checking. Yeah, I'm comfortable using `\n`, as long as it works on all platforms :)

mpMelnikov:
@joelmartinez I've pushed the requested changes. (Not sure whether I should write this comment or you get a notification when I push to the pull requested branch)

mdoc/mdoc.csproj
```diff
@@ -55,6 +55,8 @@
     <Compile Include="Mono.Documentation\Updater\Formatters\StandardFlagsEnumFormatter.cs" />
     <Compile Include="Mono.Documentation\Updater\Formatters\DefaultAttributeValueFormatter.cs" />
     <Compile Include="Mono.Documentation\Updater\Formatters\AttributeValueFormatter.cs" />
+    <Compile Include="Mono.Documentation\Updater\Formatters\VBFullMemberFormatter.cs" />
+    <Compile Include="Mono.Documentation\Updater\Formatters\VbMemberFormatter.cs" />
```
dend: @mpMelnikov just from a naming perspective, can we make sure that the file names follow a standard casing approach, e.g. if we have `VBFullMemberFormatter`, then we also call the adjacent file `VBMemberFormatter`

cc: @joelmartinez

mpMelnikov: @dend Yes, you're right, I'll fix it.

joelmartinez:
build

mdoc/mdoc.csproj
joelmartinez: Good catch @dend

joelmartinez:
@mpMelnikov yes, thank you for commenting ... I do not get a notification when you push new changes, so the comment is helpful in drawing attention to it :)

joelmartinez:
@mpMelnikov just fyi, that in the process of working on #106, I've discovered something that will likely affect all the other formatters (including the vb, F#, c++, and javascript ones that are in development). I plan on finishing and merging #106 tomorrow, at which point I will post on this PR with further instructions on how to incorporate the change :) 

/fyi @katebialiatka, @asokolovskiy15; I couldn't find Eugene's GH account, but please let him know as well.

joelmartinez:
Ok, I've just published the latest release ... so if you rebase this branch, you will find that I've added a new protected method to `MemberFormatter`. You can see how it's used in this commit on the C# formatter:  
https://github.com/mono/api-doc-tools/commit/6ae59208caf3d2dfe6333897cf4062a30d82a87c#diff-a91098a51b117f5bb4d99be3e7c68686R646 

TL;DR every formatter (I made the same change to the ILAsm formatter) needs to check the `IsPublicEII` method in order to properly format a signature for explicitly implemented interface events.

mpMelnikov:
@joelmartinez 
I've pushed EII events changes for VB.NET. It's a separate commit, at the moment. I will squash all commits when you tell us that you are ready to merge this pull request.
There seems to be a little issue with interface events in C# formatter. Now it generates:
```
  <TypeSignature Language="C#" Value="public interface MyPublicInterface" />
...
      <MemberSignature Language="C#" Value="public event EventHandler&lt;string&gt; PublicEvent;" />
```
There shouldn't be `public` modifier before the event declared in the interface. I've added such a check for VB.NET formatter:
```
            if (AppendVisibility(buf, e.AddMethod).Length == 0 && !isPublicEII)
            {
                return null;
            }
            if (e.DeclaringType.IsInterface) // There is no access modifiers in interfaces
            {
                buf.Clear();
            }
```

joelmartinez:
That's a really good catch! I'm going to make this change as well ...

joelmartinez:
@mpMelnikov this looks like it is ready to merge. Please squash commits and I'll get it rebased onto `master` 🎉

PR #143, opened 2017-10-25
mdoc: Preserving remarks created post-factum on import

### LINKED ISSUE #97: [mdoc] Preserving remarks created post-factum on import
One of the issues that we are running when we are importing documentation from docXML is that any remarks that were created in the post-production ECMAXML are erased in favor of the imported ones. Here is the scenario.

1. We onboard a new set of DLLs and import Roslyn-generated docXMLs, giving us a new batch of ECMAXML files.
2. Writers edit the ECMAXML remarks to add code examples.
3. A new version of a packages comes out, so we re-run the `mdoc` to import the new information.
4. The import will overwrite the remarks already in ECMAXMLs in favor of those that are in the files already.

This issue was brought up by @mmacy with the Azure Storage libraries, that rely heavily on remarks and user-created samples.

asokolovskiy15:
Added attribute "overwrite" for the following elements: remarks, summary, value, returns, param, typeparam, altmember, exception, permission
Closes #97

### DIFF
--- a/mdoc/Mono.Documentation/Updater/DocUtils.cs
+++ b/mdoc/Mono.Documentation/Updater/DocUtils.cs
@@ -200,6 +200,13 @@ namespace Mono.Documentation.Updater
                     baseRef.FullName == "System.MulticastDelegate";
         }
 
+        public static bool NeedsOverwrite(XmlElement element)
+        {
+            return element != null &&
+                   !(element.HasAttribute("overwrite") &&
+                    element.Attributes["overwrite"].Value.Equals("false", StringComparison.InvariantCultureIgnoreCase));
+        }
+
         public static List<TypeReference> GetDeclaringTypes (TypeReference type)
         {
             List<TypeReference> decls = new List<TypeReference> ();
--- a/mdoc/Mono.Documentation/Updater/MsxdocDocumentationImporter.cs
+++ b/mdoc/Mono.Documentation/Updater/MsxdocDocumentationImporter.cs
@@ -44,15 +44,19 @@ namespace Mono.Documentation.Updater
 
             XmlElement e = info.Node;
 
-            if (elem.SelectSingleNode ("summary") != null)
-                MDocUpdater.ClearElement (e, "summary");
-            if (elem.SelectSingleNode ("remarks") != null)
-                MDocUpdater.ClearElement (e, "remarks");
-            if (elem.SelectSingleNode ("value") != null || elem.SelectSingleNode ("returns") != null)
+            if (elem.SelectSingleNode("summary") != null && DocUtils.NeedsOverwrite(e["summary"]))
+                MDocUpdater.ClearElement(e, "summary");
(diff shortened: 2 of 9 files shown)


joelmartinez:
@asokolovskiy15 please resolve the branch conflicts :)

mdoc/Mono.Documentation/Updater/DocUtils.cs
```diff
@@ -200,6 +200,13 @@ public static bool IsDelegate (TypeDefinition type)
                     baseRef.FullName == "System.MulticastDelegate";
         }
 
+        public static bool NeedOverwrite(XmlElement element)
+        {
+            return element != null &&
+                   (!element.HasAttribute("overwrite") ||
+                    element.Attributes["overwrite"].Value != "false");
```
joelmartinez: Please change this to use a case insensitive check, that way, in case a user ever uses `"False"`, it will still work

mdoc/Mono.Documentation/Updater/DocUtils.cs
```diff
@@ -200,6 +200,13 @@ public static bool IsDelegate (TypeDefinition type)
                     baseRef.FullName == "System.MulticastDelegate";
         }
 
+        public static bool NeedOverwrite(XmlElement element)
```
joelmartinez: Very minor, but please change the name of this member to `NeedsOverwrite` (note the `s`) ... it makes the code easier to read, grammatically speaking.

mdoc/Mono.Documentation/Updater/MsxdocDocumentationImporter.cs
```diff
@@ -75,8 +83,12 @@ public override void ImportDocumentation (DocsNodeInfo info)
                     case "returns":
                         {
                             XmlElement v = e.OwnerDocument.CreateElement (info.ReturnIsReturn ? "returns" : "value");
-                            v.InnerXml = child.InnerXml;
-                            e.AppendChild (v);
+                            XmlElement p2 = (XmlElement)e.SelectSingleNode(child.Name);
+                            if (p2 == null)
```
joelmartinez: Unless I'm reading the code wrong (a distinct possibility :) ), should this `case` also contain a call to `NeedsOverwrite`?

asokolovskiy15: In this case p2 is already null, because we have cleared this node here:
 ```
 if (DocUtils.NeedsOverwrite(e["value"]))
    MDocUpdater.ClearElement(e, "value");
```
At the moment, `NeedsOverwrite` returns false for null values, so call `NeedsOverwrite` here will break the logic.

joelmartinez: Oh, I see ... carry on :)

joelmartinez:
build

PR #145, opened 2017-10-27
mdoc: EII events are now documented

### LINKED ISSUE #106: EII events should not be treated as private

joelmartinez:
Closes #106 ... also bumps version to 5.2

### DIFF
--- a/mdoc/Consts.cs
+++ b/mdoc/Consts.cs
@@ -3,6 +3,6 @@ namespace Mono.Documentation
 {
 	public static class Consts
 	{
-		public static string MonoVersion = "5.1.1.0";
+		public static string MonoVersion = "5.2.0.0";
 	}
 }
--- a/mdoc/Makefile
+++ b/mdoc/Makefile
@@ -540,4 +540,4 @@ check-doc-tools-update: check-monodocer-since-update \
 	check-mdoc-validate-update 
 
 check: nunit check-doc-tools 
-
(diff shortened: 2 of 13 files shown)


joelmartinez:
build

Issue #108, opened 2017-08-11
[mdoc] Support for F#-specific entities


dend:
Before we get to the actual implementation details, would be great if @cartermp and @dsyme could outline the general F# language conventions that need to be accounted for during reflection.

joelmartinez:
I'm really excited about improving support for F# in mdoc tooling, cause I <3 F# :) Just for tracking purposes, we have another issue, #10, which requests additional signatures. However, we will use this issue for the specific purposes of defining scope of work and requirements. 

Implementation note: we will need to write a new `MemberFormatter` class, and add it to the [other formatters](https://github.com/mono/api-doc-tools/blob/master/mdoc/Mono.Documentation/monodocer.cs#L157-L165). Additionally, we'll have to decide whether to add this by default, or put it behind a flag so users can opt-in. I'm _leaning_ towards adding it by default going forward .. but am open to discussion on the matter.

Looking forward to getting the details :)

dsyme:
Hey @dend, @joelmartinez, here's an amail I sent to Den internally outlining some of the issues I saw:

The overall problem is that the generator is treating F# libraries as if they are just vanilla .NET/IL libraries for consumption from C#. This means we see the compiled form of F#-specific constructs. 
 That’s an OK starting point, but we really do need to modify the generator to present information in a way that’s suitable and appropriate for the F# programmer.   Specifically the doc generator should be modified to cope with a handful of F# specific constructs and names.  Once you do this everything will be much better   If the generator is open source we may be able to help with the changes.
 
* ``FSharpFunc<A,B>`` should be generated as the F# function type notation ``A -> B``.    Implement by looking for type name ``Microsoft.FSharp.Core.FSharpFunc`2``
 
* Basically every construct with an attribute [CompilationRepresenationAttribute](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/core.compilationrepresentationattribute-class-%5Bfsharp%5D) will likely need to be treated as somewhat special.  For example, curried functions should be detected and presented with the correct type signature. These have a [CompilationArgumentsCountAttribute](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/core.compilationargumentcountsattribute-class-%5bfsharp%5d) telling you the sequencing of curried arguments. Examples are List.map and Array.fold.
 
[...]

PR #146, opened 2017-10-27
mdoc: F# signatures support (#108)

mpMelnikov:
Added FSharpFullMemberFormatter (there are a lot of TODOs now)
Added FSharpConsoleApplication - it's a sample project which is used to test F# formatters. In the end, the solution will be deleted and its source code will be used in integration tests
Changed DocumentationEnumeration. It was enumerating nested types twice
Changed MDocUpdated. Now it works with the names which contain apostrophes

@joelmartinez It's an interim commit. If you have some remarks or questions, I'm ready to discuss.


joelmartinez:
Nice :) How are you thinking of structuring the integration tests ... are you going to use the F# command line compiler? Would it be easier to just keep a `.fsproj` and then use msbuild to compile it in the makefile? I'm open to whatever solution works best

joelmartinez:
Another thing, please refer to this comment (https://github.com/mono/api-doc-tools/pull/142#issuecomment-340064916), for a small additional requirement based on the latest 5.2 release.

mpMelnikov:
> How are you thinking of structuring the integration tests

@joelmartinez I've tried both variants with Ubuntu on Dockets and technically both of them work fine. I suppose we can start with a project compiled with msbuild, and compare results with 'diff'. In this case, we can easily notice which signatures are incorrect in the output of 'diff'. But if we find out some reasons that we need separated tests, we can always divide this project into parts.

joelmartinez:
Ideally, the **bulk** of the tests would be in the form of unit tests ... similar to the existing formatter tests. That should make it simple to exercise all of the significant features. Then one or two integration tests from compiled F# assemblies. I'd definitely like to avoid a docker dependency for the CI agents.

mpMelnikov:
@joelmartinez 
Ok. I've got your idea. Of course, there is no any dependency on Docker. The only new dependency will be `apt-get install fsharp`. We will have to add it to [README](https://github.com/mono/api-doc-tools/blob/master/README.md) for integration tests.

joelmartinez:
@mpMelnikov for my own purposes, I added the `work-in-progress` tag ... I'll remove it when you tell me that it's ready to review :)

mpMelnikov:
@joelmartinez We are ready for review. I haven't squashed commits yet, because there probably will be more commits after you review it.

joelmartinez:
@mpMelnikov excellent news! ok, I will start reviewing this today :)

mdoc/Mono.Documentation/mdoc.cs
```diff
@@ -16,19 +16,20 @@ class MDoc {
 		private static void Main (string[] args)
 		{
 			MDoc d = new MDoc ();
-			try {
+//			try {
```
joelmartinez: Exception handling is commented out

mdoc/Mono.Documentation/mdoc.cs
```diff
@@ -16,19 +16,20 @@ class MDoc {
...
+//				if (debug) {
+//					Console.Error.WriteLine ("mdoc: {0}", e.ToString ());
+//				}
+//				else {
+//					Console.Error.WriteLine ("mdoc: {0}", e.Message);
+//				}
+//				Console.Error.WriteLine ("See `mdoc help' for more information or use --debug to diagnose.");
+//				Environment.ExitCode = 1;
+//			}
+		    Console.ReadKey();
```
joelmartinez: Requires manual user interaction

mpMelnikov:
@joelmartinez Oh, my fault. I've just committed the fix.

mdoc/Mono.Documentation/Updater/DocumentationEnumerator.cs
```diff
@@ -27,8 +27,9 @@ protected IEnumerable<TypeDefinition> GetDocumentationTypes (AssemblyDefinition
                 if (seen != null && seen.Contains (type.FullName))
                     continue;
                 yield return type;
-                foreach (TypeDefinition nested in type.NestedTypes)
-                    yield return nested;
+                // because of this code we get nested classes twice
+//                foreach (TypeDefinition nested in type.NestedTypes)
```
joelmartinez: Commenting out code like this makes me a bit worried because it _seems_ unrelated to the other changes.

- Was there another change added that compensates for this code being commented out?
- Why weren't we seeing double nested classes before if this code was present in the codebase?
- Do we not have any nested types in our integration tests?
    - just checked, yeah, we do in `DocTest.cs` ... for example, `Widget.NestedClass`.

I'd like a bit more investigation around this code before removing it ... does it only affect the F# nested types? What happens if you add it back ... in particular, which integration tests fail. What happens if you comment out this code in the `master` branch and then run the integration tests?

mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
```diff
@@ -30,6 +30,7 @@ protected override StringBuilder AppendNamespace (StringBuilder buf, TypeReferen
         protected virtual string GetCSharpType (string t)
         {
             // make sure there are no modifiers in the type string (add them back before returning)
+            // TODO: Do we really need this code? When does it work? What is the modifier?
```
joelmartinez: yes, this is necessary. it's related to `modreq` and `modopt` modifiers ... mainly added by c++ compilers, but are not supported in any way in C# currently. You can remove this comment :)

mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
```diff
@@ -484,33 +485,27 @@ private StringBuilder AppendParameter (StringBuilder buf, ParameterDefinition pa
...
-                     (!method.IsPrivate && !method.IsAssembly && !method.IsFamilyAndAssembly)))
-                get_visible = AppendVisibility (new StringBuilder (), method).ToString ();
-            string set_visible = null;
-            if ((method = property.SetMethod) != null &&
-                    (DocUtils.IsExplicitlyImplemented (method) ||
-                     (!method.IsPrivate && !method.IsAssembly && !method.IsFamilyAndAssembly)))
-                set_visible = AppendVisibility (new StringBuilder (), method).ToString ();
-
-            if ((set_visible == null) && (get_visible == null))
+            string getVisible = null;
```
joelmartinez: There doesn't really seem to have been a reason to change the `get_visible` variable names, and indeed, their usage further down in the code was not changed. Please revert this change, as it just adds noise to the PR/Diff without any real change to functionality.

mdoc/Mono.Documentation/Updater/Formatters/FSharpFullMemberFormatter.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Text;
+using Mono.Cecil;
+
+namespace Mono.Documentation.Updater
+{
+    public class FSharpFullMemberFormatter : FSharpFormatter
+    {
+        public override string Language => "F#";
```
joelmartinez: This should be using `Consts.FSharp`

mdoc/Mono.Documentation/Updater/Formatters/FSharpFormatter.cs
```diff
@@ -0,0 +1,1084 @@
...
+            buf.Append(GetTypeName(type));
+            buf.Append(" = ");
+
+            if (IsRecord(type))
+            {
+                buf.Append("{}");
+                return buf.ToString();
+            }
+
+            if (IsDescriminatedUnion(type))
```
joelmartinez: spelling error, the name of the function should be `IsDiscriminatedUnion`

mdoc/Mono.Documentation/Updater/Formatters/FSharpFormatter.cs
```diff
@@ -0,0 +1,1084 @@
...
+        }
+
+        protected override StringBuilder AppendRefTypeName(StringBuilder buf, TypeReference type, DynamicParserContext context)
+        {
+            return buf;
+        }
+
+        protected override string GetFinalizerName(MethodDefinition method)
+        {
+            throw new NotSupportedException("No finalizers in F#");
```
joelmartinez: Should this throw an exception? if we try to process an assembly that contains a finalizer, is the `mdoc` process going to error out? I would prefer the formatter to return false from `IsSupported` on a finalizer member, rather than throwing an exception here ...

In fact, I'd like to see a finalizer as part of the integration tests to test this very scenario for the F# signatures.

mdoc/Mono.Documentation/Updater/Formatters/FSharpFormatter.cs
```diff
@@ -0,0 +1,1084 @@
...
+        protected override StringBuilder AppendPointerTypeName(StringBuilder buf, TypeReference type, DynamicParserContext context)
+        {
+            TypeSpecification spec = type as TypeSpecification;
+            buf.Append("nativeptr<");
+            _AppendTypeName(buf, spec != null ? spec.ElementType : type.GetElementType(), context);
+            buf.Append(">");
+            return buf;
+        }
+
+        private static string GetTab()
```
joelmartinez: It would _probably_ be a bit cleaner to remove this method, and simply add a const field to the `Consts` class ... say, `Consts.Tab`. That way, it can be used any other place where it's needed, rather than calling a method that returns a new instance of  a string.

mdoc/Mono.Documentation/Updater/DocumentationEnumerator.cs
mpMelnikov: Before answering the questions, I'd like to describe this issue in general.
When I mention double nested classes I mean this method returns nested classes twice, they don't appear twice in the result XML file. The reason why this behavior is hidden is MDocUpdater class. Usually, it processes double values correctly, first occurrence is processed as a new value, and the second occurrence refreshes already generated nodes with the same value. So, it doesn't affect the result XML, but there are excess operations. 
In F# I encountered this issue because there are extension methods in a nested class in the integration test. In this case, MDocUpdater doesn't process twice iterated members correctly.

> Was there another change added that compensates for this code being commented out?
> Why weren't we seeing double nested classes before if this code was present in the codebase?

There were no other changes to compensate this one. We weren't seeing double nested classes because this case is handled by MDocUpdater class as I described above.

> Do we not have any nested types in our integration tests? just checked, yeah, we do in DocTest.cs ... for example, Widget.NestedClass.

> does it only affect the F# nested types? 

It affects all input assemblies and formatters.

> What happens if you add it back ... in particular, which integration tests fail. 
 
check-monodocer-fsharp fails with "Failed to compare two elements in the array." The issue appears because there are extension methods in a nested class.

> What happens if you comment out this code in the master branch and then run the integration tests?

Integration tests work as expected.

mdoc/Mono.Documentation/Updater/Formatters/FSharpFormatter.cs
mpMelnikov: Removed this method. It was never called for F#

mdoc/Mono.Documentation/Updater/DocumentationEnumerator.cs
joelmartinez: Ahh I see ... excellent work rooting out the erroneous code :) Please remove it from the code entirely, as it doesn't need to stay commented out.

joelmartinez:
@mpMelnikov oh, also I should have mentioned ... please make sure to rebase when you squash all the commits. There was a small change to the makefile on another PR that was already merged into `master`, that is conflicting with this PR :)

mpMelnikov:
@joelmartinez Will it be ok if I rebase and squash commits as we are ready to merge it into master? Or is it better if I do it sooner?

joelmartinez:
@mpMelnikov oh it's totally fine if you wait ... more important just to rebase now rather than later, to resolve the conflict, and also make sure there aren't any side effects from the recently merged change (doubtful as it was a relatively small change, but still important to check) 😄

joelmartinez:
@mpMelnikov Going through the generated/integration tests code review ... have one comment so far

- Please move the`mdoc.Test.FSharp` project/folder *into* `mdoc/mdoc.Test` folder. You'll notice that I already have a `cppcli` folder in there for c++ test data. /cc @katebialiatka ... as you're working on tests for your project as well, please note this feedback.

mdoc.Test.FSharp/Delegates.fs
```diff
@@ -0,0 +1,119 @@
...
+// Use 'typ * ... * typ -> typ' for multi-argument delegates, 
+// and 'typ -> (typ -> typ)' for delegates returning function values.	
+
+type Delegate1 = delegate of (int * int) -> int// Delegate1 works with tuple arguments.
+type Delegate2 = delegate of int * int -> int // Delegate2 works with curried arguments.
+type Delegate3 = delegate of int * char -> string
+type Delegate4 = delegate of int -> (int -> char)
+type Delegate5 = delegate of int -> (int -> char -> string)
+type Delegate6 = delegate of (int -> float) -> char
+type Delegate7 = delegate of (int -> char -> string) -> float
```
joelmartinez: I find it interesting that the generated signature for this type is

```fsharp
type Delegates.Delegate7 = delegate of (int -> char -> string) -> double
```

AFAIK, F# supports both `double` and `float` ... so it's curious that the type reported here is `System.Double` for the return type. Can you look into what might be happening here?

mdoc.Test.FSharp/DiscriminatedUnions.fs
```diff
@@ -0,0 +1,18 @@
+﻿module DiscriminatedUnions
+
+//The preceding code declares a discriminated union Shape, 
+// which can have values of any of three cases: Rectangle, Circle, and Prism. 
+// Each case has a different set of fields.
+type Shape =
```
joelmartinez: Upon casual inspection, it seems like a lot of the samples in this section come from the microsoft docs ... for example, this one can be found here:
https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/discriminated-unions#remarks

Perhaps this is a question for @dend ... but as that content is available on a creative commons/attribution license, do we need to include any kind of attribution in these source files?

mdoc.Test.FSharp/Delegates.fs
mpMelnikov: @joelmartinez yes, you're right, F# supports `double` and `float`, but both of them are `System.Double` in IL code ([type match table](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/primitive-types)). In this case, I've decided to generate `double` cause it seems to be closer to the real IL code. There is a similar issue with `float32` and `single` which really are `System.Single`, I generate `single` for them.

mdoc.Test.FSharp/Functions.fs
```diff
@@ -0,0 +1,27 @@
...
+// Functions in F# can be composed from other functions. 
+// The composition of two functions function1 and function2 is another function that represents the application of function1 followed the application of function2:
+let function1 x = x + 1
+let function2 x2 = x2 * 2
+let function3 (x3) = x3 * 2
+let function4 x4 y4 = x4 + y4
+let function5 (x5, y5) = x5 + y5
+let function6 (x6, y6) = ()
+let function7 x7 (y7, z7) = ()
+let function8 x8 y8 z8 = ()
```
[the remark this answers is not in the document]
mpMelnikov: @joelmartinez Well, it's a really interesting question. I'll investigate this issue more thoroughly and will prepare an answer (or source code changes).

mpMelnikov:
@joelmartinez 
>... Do the existing MSDN docs include the function8<'a,'b,'c> portion?
 
Yes, you are right, I should remove these "explicit type parameters" from methods signature.
For F#, we show all functions and methods as properties of function type, but properties can't have explicit type parameters:
```
module Functions

let function13<'a> (a:'a) = ()

type TestFunction() =
    member this.f13 : 'a -> unit = function13
    member this.f13<'a> : 'a -> unit = function13// Error FS0671 A property cannot have explicit type parameters. Consider using a method instead.
```

An issue is constraints because in this case they can't be shown in F#-syntax manner. That's why MSDN and VS display a remark `(requires ...)`
image
```
type Constraints.Class1<'T (requires 'T :> Exception)> = class
```
So, I remade constraints signatures and tests as well.

> are there any particular areas that you would like additional/specific scrutiny on?

Well, It's a difficult question:) There were many interesting tasks. I'm a little worried about the following moments:
1. I display `this.` modifier before non-static type members. It's how they are displayed in [this MSDN article](https://msdn.microsoft.com/visualfsharpdocs/conceptual/control.mailboxprocessor%5b%27msg%5d-class-%5bfsharp%5d), for example. Moreover, it can be compiled successfully, but Visual Studio displays name of the type instead of `this`.
2. Constraints. We've already discussed many of them are not presented in IL.

mpMelnikov:
3. Implementation of generic interfaces. [In MSDN](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/collections.map%5b'key,'value%5d-class-%5bfsharp%5d), we don't see mentions of generic arguments, I've implemented the same functionality:
```
            foreach (var interfaceImplementation in type.Interfaces)
            {
                if (type.IsValueType
                    && ignoredValueTypeInterfaces.Any(i => interfaceImplementation.InterfaceType.FullName.StartsWith(i)))
                    continue;
                buf.Append($"{GetLineEnding()}{Consts.Tab}interface ");
                AppendTypeName(buf, interfaceImplementation.InterfaceType.Name);
                // We can use full name ( GetTypeName(interfaceImplementation.InterfaceType) ), 
                // but MSDN shows only names
            }
```
But I'm not sure whether it's a good choice.

joelmartinez:
@mpMelnikov 

- on the use of `this` ... hmmmm. I think I lean more towards supporting the use of `this` in this instance, because a reader could copy and paste that into their own type and have it work (if they want to have a member that uses the same signature, for example). I say leave it as is :)
- Your note on constraints is duly noted ... we discussed this, and for the time being this is acceptable. Because this information is not present in the assembly, the only way to add it would be to refer to the source code, which is not something that we're ready to do at this stage (if ever).
- For the generic arguments on interfaces ... Please add these to the listing. Not sure why it's not listed on MSDN, but this loses a lot of information if we use only the name.

joelmartinez:
Once you add the generic arguments, I think it's ready for the commits to be squashed and merging 🎉

mpMelnikov:
@joelmartinez It's all done. There was a really little change to add the generic arguments on interfaces (one line of code + tests), so I've squashed commits at once.

cartermp:
Thank you @mpMelnikov!

mpMelnikov:
@joelmartinez @cartermp you are welcome and thank you, guys! It was a really interesting task.

PR #148, opened 2017-10-31
mdoc: updates for WSL

joelmartinez:
With these changes, you can now run the make build on windows, via WSL. I've made some small changes to the tests to account for path case sensitivity, updated the README with information on upgrading to the latest version of WSL, and lastly, just added some additional informative output to help in troubleshooting the loading of frameworks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,16 +33,18 @@ If you want to compile in debug mode: `make all CONFIGURATION=Debug`
 
 ### WSL
 
-On the windows subsystem for linux, there is unfortunately a defect that causes difficulty making https calls, which in turn doesn't allow the `prepare` make target to complete (nuget fails). You will see an error that says:
+Make sure you are at least on 16.04 (Xenial). You can check which version is currently installed by running the command `lsb_release -a`.
 
-```
-Unable to load the service index for source https://api.nuget.org/v3/index.json.
-  An error occurred while sending the request
-  Error: ConnectFailure (Value does not fall within the expected range.)
-  Value does not fall within the expected range.
-```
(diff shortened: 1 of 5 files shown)


joelmartinez:
@mpMelnikov looks like something is strange with the analytics mode. I changed one of the expected results in order for this to pass locally, and on closer inspection, it seems that the change I made is correct. 

```
diff -rup Test/expected_fx_remove_statistics.txt Test/actual_statistics.txt
--- Test/expected_fx_remove_statistics.txt	2017-10-31 10:57:12.000000000 -0400
+++ Test/actual_statistics.txt	2017-10-31 10:58:17.000000000 -0400
@@ -1,7 +1,7 @@
 Framework: one
 --------
 Types Added: 0
-Types Removed: 2
+Types Removed: 1
 Types Total: 2
 
 Namespaces Added: 0
```

To be specific, the integration test `check-monodocer-fx-statistics-remove` first creates a docset using `Test/DocTest-DropNS-unified-deletetest.dll`, and then does a subsequent run of mdoc using `-statistics` against `Test/DocTest-DropNS-unified.dll`. This version of the assembly does in fact, remove two types: `WillDelete`, and `OnlyInMulti`.

What's strange is that locally (on both WSL, using the version on this branch), and on the Mac, I get this result ... while it fails on the CI server. Can you look and see what might be the cause of this issue? Thanks!

mpMelnikov:
@joelmartinez 
I hope it's fixed here: #149. I've updated the integration test. But it seems that Jenkins still runs tests without the last changes and fails.

PR #167, opened 2017-11-28
mdoc: Support for Javascript signatures

### LINKED ISSUE #133: [mdoc] Support for JavaScript signatures
Support API entity signatures in JavaScript.

### Exception 1:

For the V1 Pri1 implementation, we will not implement custom “retrievers”.  If a non-static class doesn’t have a public constructor (in other words, it is not possible to automatically determine the call to instantiate an instance of the class), the Javascript syntax should either:

* show a standard disclaimer such as:
“This class does not provide a public constructor” or
“See the remarks section for information on obtaining an instance of this class”
* Give a degenerate declarative syntax, such as simply: “Windows.System.FolderLauncherOptions” for the FolderLauncherOptions class.
The specific solution to use here is still TBD.  If you’re blocked, pick A1 for now.
We will investigate whether a Pri 2 feature to modify the syntax block with custom syntax is necessary.

## How to determine if an API supports JavaScript

Use the `WebHostHidden` attribute. If `WebHostHidden` is present, the API doesn’t support JavaScript. None of the APIs in the “XAML” namespaces support JavaScript. 

#### Usage (get)

```
var [property value type, camel-cased] = [typename, camel-cased].[property name, camel-cased];
```

#### Usage (set)

```
[typename, camel-cased].[property name, camel-cased] = [property value type, camel-cased];
```

#### Usage (get)

```
var [property value type, camel-cased] = [typename].[property name, camel-cased];
```

#### Usage (set): 

```
[typename].[property name, camel-cased] = [property value type, camel-cased];
```

**Note:** property value might vary between JavaScript, C# and C++. Use the same filtering behavior that is used for C# and C++. 

#### FolderLauncherOptions.DesiredRemainingView property 

Specifies how much space on the screen to leave for the calling app when launching File Explorer. The calling app can take up more or less space than File Explorer, or File Explorer can fill the entire screen.
##### Syntax

**JavaScript (usage)**
```
var viewSizePreference = folderLauncherOptions.desiredRemainingView;
folderLauncherOptions.desiredRemainingView = desiredRemainingView;
```

**C#**

```
public ViewSizePreference DesiredRemainingView { get; set; }
```

**VB**
```
Public Property DesiredRemainingView As ViewSizePreference
```

**C++**

```
public:
property ViewSizePreference DesiredRemainingView { 
   ViewSizePreference get();
   void set (ViewSizePreference value);
}
```

**Property value**
Type: `ViewSizePreference`
One of the enumeration values that specifies how much space on the screen to leave for the calling app when launching File Explorer. 

### Enums

Usage syntax

```
var value = [fully qualified type].[camel-cased name of first enum value];
```

**Note:** member names for JavaScript should be camel-cased. 

##### ProcessorArchitecture enumeration

Specifies the processor architecture supported by an app.

**Syntax**

**JavaScript**

```
var value = Windows.System.ProcessorArchitecture.x86;
```

**C#**

```
public enum ProcessorArchitecture
```

**VB**

```
Public Enum ProcessorArchitecture
```

**C++**

```
public enum class ProcessorArchitecture
```

**Members**
The ProcessorArchitecture enumeration has these members.

|Member|	Value|Description
|-----|------|-----|
|X86 \| x86 |0|The x86 processor architecture.|
|Arm \| arm|5|The ARM processor architecture.|
|X64 \| x64|9|The x64 processor architecture.|
|Neutral \| neutral|11|A neutral processor architecture.|
|Unknown \| unknown|65535|An unknown processor architecture|

### Async Methods

(For static methods, use the fully-qualified class name. For non-static, use the class name without qualification and camel-cased.)

Usage: 

```
[fully qualified type name].[camel-cased method name]([parameter names]).done( /* Your success and error handlers */ );
```

**Note:** parameter and return types might be different from C#. Need to provide new type mapping. 

**Launcher.FindUriSchemeHandlersAsync(String, LaunchQuerySupportType) method**

Enumerate the scheme handlers on the device.

**Syntax**

**JavaScript (usage)**

```
Windows.System.Launcher.findUriSchemeHandlersAsync(scheme, launchQuerySupportType).done( /* Your success and error handlers */ );
```

**C#**

```
public static IAsyncOperation<IReadOnlyList> FindUriSchemeHandlersAsync(
  string scheme, 
  LaunchQuerySupportType launchQuerySupportType
)
```

**VB**

```
Public Shared Function FindUriSchemeHandlersAsync(
  scheme As String,  
  launchQuerySupportType As LaunchQuerySupportType 
) As IAsyncOperation(Of IReadOnlyList)
```

**C++**

```
public:
static IAsyncOperation<IVectorView>^ FindUriSchemeHandlersAsync(
  String^ scheme, 
  LaunchQuerySupportType launchQuerySupportType
)
```

**Parameters**

`scheme` 

Type: `String [JavaScript] | System.String [.NET] | Platform::String [C++]`

The scheme name that you find to find handlers for. For example, "ms-lens".

### Methods

**Non-static**

**Usage (not void)**

```
var [return type, camel-cased] = [class-name, camel-cased].[method-name, camel-cased]([parameter name list]);
```

**Usage (void)**

```
[class-name, camel-cased].[method-name, camel cased]([parameter name list]);
```

**Static**

**Usage (not void)**

```
var [return type, camel-cased] = [fully-qualified class-name].[method-name, camel-cased]([parameter name list]);
```

**Usage (void)**

```
[fully-qualified class name].[method-name, camel cased]([parameter name list]);
```

**Example**

**MemoryManager.TrySetAppMemoryUsageLimit method**

Tries to set a specific memory cap for the current app or task. In cases where memory caps are shared between foreground and background components, any difference between the default cap and the new request will be assigned to the other component.

**Syntax**

**JavaScript (usage)**

```
var boolean = Windows.System.MemoryManager.trySetAppMemoryUsageLimit(value);
```

**C#**

```
public static bool TrySetAppMemoryUsageLimit(
  ulong value
)
```

**VB**

```
Public Shared Function TrySetAppMemoryUsageLimit(
  value As ULong 
) As Boolean
```

**C++**

```
public:
static bool TrySetAppMemoryUsageLimit(
  unsigned long long value
)
```

**Parameters**

`value`
 
Type: `Number [JavaScript] | System.UInt64 [.NET] | uint64 [C++]`

The new memory limit to set.

**Return value**

Type: `Boolean [JavaScript] | System.Boolean [.NET] | Platform::Boolean [C++]`

True if the operation succeeded; otherwise, false.

### Events
Usage: 

```
function on[EventName](eventArgs){/* Your code */}
[class name, camel cased].addEventListener(“[event name, camel cased]”, on[EventName]); 
[class name, camel cased].removeEventListener(“[event name, camel cased]”, on[EventName]);
```

•	If the event supports property syntax (it has a corresponding EventName property), add this:

```
- or -
[class name, camel cased].on[event name, all lower-case] = on[EventName];
```

**Example**

**MemoryManager.AppMemoryUsageIncreased event**

Raised when the app's memory consumption has increased to a higher value in the AppMemoryUsageLevel enumeration.

**Syntax**

**JavaScript (usage)**

```
function onAppMemoryUsageIncreased(eventArgs) { /* Your code */ }
 
// addEventListener syntax
memoryManager.addEventListener("appmemoryusageincreased", onAppMemoryUsageIncreased);
memoryManager.removeEventListener("appmemoryusageincreased", onAppMemoryUsageIncreased);
```
- or -
```
memoryManager.onappmemoryusageincreased = onAppMemoryUsageIncreased;
```

**C#**

```
public static event EventHandler<object> AppMemoryUsageIncreased
```

**VB**

```
Public Shared Event AppMemoryUsageIncreased As EventHandler(Of Object)
```

**C++**

```
public:
static event EventHandler<object>^ AppMemoryUsageIncreased {
   Windows::Foundation::EventRegistrationToken add(EventHandler<object>^ value);
   void remove(Windows::Foundation::EventRegistrationToken token);
}
```

**Event information**

**Delegate** 

EventHandler [JavaScript] | System.EventHandler<Object> [.NET] | Windows.Foundation.EventHandler<Object> [C++]

### Structures

**Usage**

```
var [struct name, camel cased] = {
	[fieldname, came cased] : /* Your value */
}
```

**Example**

```
var rect = {
    height : /* Your value */,
    width: /* Your value */,
 
    x: /* Your value */, 
    y: /* Your value */
}
```

https://docs.microsoft.com/en-us/uwp/api/windows.foundation.rect  

### Delegates

**Usage**

```
var [delegateName, camel-cased]Handler = function([parameter name list, camel cased]){
  /* Your code */
}
```

**Example:**

```
var asyncActionCompletedHandler = function(asyncInfo, asyncStatus) {
 /* Your code */
}
```

https://docs.microsoft.com/en-us/uwp/api/windows.foundation.asyncactioncompletedhandler 

### Interfaces

**Usage (missing retrievers)**

```
// You can’t instantiate an interface directly in JavaScript. You can use objects that implement the interface, however. 
```

Retrievers for interfaces and types without constructors (factory pattern)
Consider adding a dedicated “retrievers” section that tell readers how to retrieve an instance of interfaces and classes that don’t have a public constructor.

mpMelnikov:
Added Javascript usage formatter
Added unit tests
Closes #133

### DIFF
--- a/mdoc/Mono.Documentation/Updater/Formatters/JsUsageFormatter.cs
+++ b/mdoc/Mono.Documentation/Updater/Formatters/JsUsageFormatter.cs
@@ -1,0 +1,246 @@
+﻿using System.Linq;
+using System.Text;
+using Mono.Cecil;
+using Mono.Documentation;
+using Mono.Documentation.Updater;
+using Mono.Documentation.Util;
+
+namespace mdoc.Mono.Documentation.Updater.Formatters
+{
+    public class JsUsageFormatter : JsFormatter
+    {
+        public override string Language => "JavaScript (usage)";
+
+        protected override string GetPropertyDeclaration(PropertyDefinition property)
+        {
+            bool getVisible = property.GetMethod != null && property.GetMethod.IsPublic;
+            bool setVisible = property.SetMethod != null && property.SetMethod.IsPublic;
+            var method = property.SetMethod ?? property.GetMethod;
+
+            // https://github.com/mono/api-doc-tools/issues/133
+            // var [property value type, camel-cased] = [typename, camel-cased].[property name, camel-cased];
+            // [typename, camel-cased].[property name, camel-cased] = [property value type, camel-cased];
+            // Static properties
+            // var [property value type, camel-cased] = [typename].[property name, camel-cased];
+            // [typename].[property name, camel-cased] = [property value type, camel-cased];
+            var buf = new StringBuilder();
+            var propertyValueType = CamelCase(GetName(property.PropertyType));
+            var propertyName = DocUtils.GetPropertyName(property);
+            var propertyNameCamelCased = CamelCase(propertyName);
+            var typeName = GetName(property.DeclaringType);
+            if (!method.IsStatic)
+                typeName = CamelCase(typeName);
+
+            if (getVisible)
+            {
+                buf.Append("var ");
+                buf.Append(propertyValueType);
+                buf.Append(" = ");
+                buf.Append(typeName);
+                buf.Append(".");
+                buf.Append(propertyNameCamelCased);
+                buf.Append(";");
+            }
+
+            if (setVisible)
+            {
+                if (getVisible)
+                    buf.Append(GetLineEnding());
+
+                buf.Append(typeName);
+                buf.Append(".");
+                buf.Append(propertyNameCamelCased);
+                buf.Append(" = ");
+                buf.Append(propertyValueType);
+                buf.Append(";");
+            }
+            return buf.ToString();
+        }
+
+        protected override string GetTypeDeclaration(TypeDefinition type)
+        {
+            var buf = new StringBuilder();
+
+            if (type.IsEnum)
+            {
+                // var value = [fully qualified type].[camel-cased name of first enum value];
+                var firstElement = type.GetMembers().Skip(1).First();//skip "__value element"
+
+                buf.Append("var value = ");
+                buf.Append(ProcessFullName(type.FullName));
+                buf.Append(".");
+                buf.Append(CamelCase(firstElement.Name));
+                return buf.ToString();
+            }
+            if (type.IsValueType)
+            {
+                //Structures: Windows Runtime structures are objects in JavaScript. 
+                // If you want to pass a Windows Runtime structure to a Windows Runtime method, 
+                // don't instantiate the structure with the new keyword. Instead, create an object 
+                // and add the relevant members and their values. The names of the members should be in camel case:
+                // SomeStruct.firstMember.
+
+                // var [struct name, camel cased] = {
+                //	[fieldname, came cased] : /* Your value */
+                buf.Append("var ");
+                buf.Append(CamelCase(GetName(type)));
+                buf.Append(" = {");
+                buf.Append(GetLineEnding());
+                buf.Append(string.Join("," + GetLineEnding(), 
+                    type.Fields.Select(i => CamelCase(i.Name) + " : /* Your value */")));
+                buf.Append(GetLineEnding());
+                buf.Append("}");
+                return buf.ToString();
+            }
+            if (DocUtils.IsDelegate(type))
+            {
+                //  var [delegateName, camel-cased]Handler = function([parameter name list, camel cased]){
+                //  /* Your code */
+                //            }
+                MethodDefinition invoke = type.GetMethod("Invoke");
+                buf.Append("var ");
+                buf.Append(CamelCase(GetName(type)));
+                buf.Append("Handler = function(");
+                AppendParameters(buf, invoke, invoke.Parameters);
+                buf.Append("){");
+                buf.Append(GetLineEnding());
+                buf.Append("/* Your code */");
+                buf.Append(GetLineEnding());
+                buf.Append("}");
+                return buf.ToString();
+            }
+            
+            var publicConstructor = GetConstructor(type);
+            return GetDeclaration(publicConstructor);
+        }
+
+        protected override string GetMethodDeclaration(MethodDefinition method)
+        {
+            var buf = new StringBuilder();
+
+            if (IsAsync(method))
+            {
+                // Async Methods
+                // (For static methods, use the fully-qualified class name. For non-static, 
+                // use the class name without qualification and camel-cased.)
+                // [fully qualified type name].[camel-cased method name]([parameter names]).done( /* Your success and error handlers */ );
+                var typeName = CamelCase(GetName(method.DeclaringType));
+                if (method.IsStatic)
+                    typeName = ProcessFullName(method.DeclaringType.FullName);
+
+                buf.Append(typeName);
+                buf.Append(".");
+                buf.Append(GetMethodName(method));
+                buf.Append("(");
+                AppendParameters(buf, method, method.Parameters);
+                buf.Append(").done( /* Your success and error handlers */ )");
+
+                return buf.ToString();
+            }
+
+            // For static and non-static method generate different signatures:
+            // 1) Non-static
+            // Usage (not void)
+            // var [return type, camel-cased] = [class-name, camel-cased].[method-name, camel-cased]([parameter name list]);
+            // Usage (void)
+            // [class-name, camel-cased].[method-name, camel cased]([parameter name list]);
+            // 2) Static
+            // Usage (not void)
+            // var [return type, camel-cased] = [fully-qualified class-name].[method-name, camel-cased]([parameter n
+            // Usage (void)
+            // [fully-qualified class name].[method-name, camel cased]([parameter name list]);
+            if (method.ReturnType != null && ProcessFullName(method.ReturnType.FullName) != Consts.VoidFullName)
+            {
+                buf.Append("var ");
+                buf.Append(CamelCase(GetName(method.ReturnType)));
+                buf.Append(" = ");
+            }
+            var className = method.IsStatic ? ProcessFullName(method.DeclaringType.FullName) : CamelCase(GetName(method.DeclaringType));
+            buf.Append(className);
+            buf.Append(".");
+            buf.Append(CamelCase(method.Name));
+            buf.Append("(");
+            AppendParameters(buf, method, method.Parameters);
+            buf.Append(")");
+
+            return buf.ToString();
+        }
+
+        protected override string GetConstructorDeclaration(MethodDefinition constructor)
+        {
+            var buf = new StringBuilder();
+
+            var typeName = GetName(constructor.DeclaringType);
+            buf.Append("var ");
+            buf.Append(CamelCase(typeName));
+            buf.Append(" = new ");
+            buf.Append(typeName);
+            buf.Append("(");
+            AppendParameters(buf, constructor, constructor.Parameters);
+            buf.Append(");");
+
+            return buf.ToString();
+        }
+
+        private bool IsAsync(MethodDefinition method)
+        {
+            return method.CustomAttributes.Any(i => i.AttributeType.FullName == "System.Runtime.CompilerServices.AsyncStateMachineAttribute");
+        }
+
+        protected override StringBuilder AppendNamespace(StringBuilder buf, TypeReference type)
+        {
+            return buf;
+        }
+
+        protected override string GetEventDeclaration(EventDefinition e)
+        {
+            // Usage:
+            //function on[EventName](eventArgs){/* Your code */}
+            //[class name, camel cased].addEventListener(“[event name, all lower-case]”, on[EventName]); 
+            //[class name, camel cased].removeEventListener(“[event name, all lower-case]”, on[EventName]);
+            // If the event supports property syntax (it has a corresponding EventName property), add this:
+            // - or -
+            //[class name, camel cased].on[event name, all lower-case] = on[EventName];
+            if (!e.AddMethod.IsPublic)
+                return null;
+
+            var className = e.AddMethod.IsStatic ? ProcessFullName(e.DeclaringType.FullName) : CamelCase(GetName(e.DeclaringType));
+            var eventName = e.Name;
+            var eventNameLowerCased = eventName.ToLower();
+            var buf = new StringBuilder();
+            buf.Append("function on");
+            buf.Append(e.Name);
+            buf.Append("(");
+            buf.Append("eventArgs");
+            buf.Append(") { /* Your code */ }");
+
+            buf.Append(GetLineEnding());
+            buf.Append(className);
+            buf.Append(".addEventListener(\"");
+            buf.Append(eventNameLowerCased);
+            buf.Append("\", on");
+            buf.Append(eventName);
+            buf.Append(");");
+
+            buf.Append(GetLineEnding());
+            buf.Append(className);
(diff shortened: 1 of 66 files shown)


mpMelnikov:
This usage formatter is based on #133 task description. I suppose, there can be some changes after we get some additional information about Javascript. 

Some remarks about implementation. For type declarations, I've implemented the following signatures.
Signature for any interface:
```
// You can’t instantiate an interface directly in JavaScript. You can use objects that implement the interface, however.
```
Signature for class with public constructors:
```
var testClassThree = new TestClassThree(i, j);
```
For types with many public constructors, I select one with maximum number of arguments. If type has public constructors with the same number of arguments I don't generate signature as this type is not supported by WinRT.

For abstract types, it was proposed in the last e-mail:
>For abstract types or types without a constructor (types that use the factory pattern, for example), you’d need to get a reference to an instance from a static method or a method of another class. 
var instanceOfFooClass = instanceOfSomeOtherClass.GetFooClass();
If we want to show the syntax, we have to figure out what returns an object of type FooClass. Which in some cases might be very complicated. For example, if you have an interface (IFoo), but the platform only has methods that return an implementation of that interface (Foo), you’d have to do a lot of processing to figure out what returns a type that can be cast as the interface.

Do we decide to show this syntax? In that case, should we iterate throw all types and members of an assembly to find methods which return a required type (or something assignable to the required type)? What if we find more than one method?

mpMelnikov:
Hi @joelmartinez!

As we found out, JavaScript can’t be compiled as .winmd libraries, so we can't create assembly on Javascript and consume it elsewhere. Declaration signatures can be consumed only by JavaScript itself. 
JavaScript is a really dynamic and weakly typed language, that's why it's hard to come up with declaration signatures on JavaScript in many cases.

In this pull request, I suggest two formatters: `JsMemberFormatter` and `JsUsageFormatter`. You can see, `JsMemberFormatter` has a really basic set of features, it generates signatures for methods and for types (signature of its constructor). As Js is dynamic and weakly typed, I can't find the way to _declare_ a property of a type, but usage samples are generated. The same about events, structs, and many other entities. I treat fields as not supported because in WinRT they can be exposed only by structures.

`JsUsageFormatter` implements what is described [here](https://docs.microsoft.com/en-us/scripting/jswinrt/using-the-windows-runtime-in-javascript) and in the issue #133.

Now `<MemberSignature>` node is not generated if there is no declaration signature. If we decide some members can have usage sample without declaration signature, something should be done with `<MemberSignature>` generation.

There were a lot of interesting facts from the universe of Js and WinRT in the e-mails, and I want to discuss some of them.

> [RP] This depends on your target platform. In UWP, it’s the same “version” of ES that is currently supported by the Edge browser.  For Windows 8.0, this was IE10, and for Windows 8.1, this was IE11.  Since Windows 10 and Edge, the browser has been evergreen, and is well into the ES2015+ versions.

For compatibility with older versions, I generate a signature of a constructor instead of `class` syntax from ES2015. Is it ok?

> Is there documentation on common language projections. Some examples of what I mean, are the base types (int and string are obvious, but what about datetime, etc.), or other "conversions" such as IAsyncAction<> exposed as a Promise.
[RP] See MSDN: Using the Windows Runtime in JavaScript, JavaScript Representation of Windows Runtime Types, and Windows Runtime DateTime and TimeSpan Representations.

Where can we replace IAsyncAction with a Promise in signatures?


> Windows Design Guidelines in general prevent concrete classes which have type parameters; only interfaces may have type parameters.  As a result, you don’t ever need to instantiate generic types.  If you need to pass an array of some object (let’s say the C++/CX corollary of Platform::Array<Platform::String^>^), you would just create a JavaScript array (e.g., []), populate it with JavaScript strings, and the projection would handle marshaling for you.

So, how does it affect signatures? Should we generate a signature and/or an usage sample for them? At the moment I treat generic types and methods as not supported both for declarations and samples. 

> [RP] Finalizer is not supported for WinRT at the ABI level.  The convention here is for WinRT objects to support the Windows::Foundation::IClosable interface, which is projected to IDisposable in .NET, to support explicit resource release.  JavaScript code can call .close() on a closable object.

Should we generate a signature of "close" method for destructors? At the moment, destructors are not supported by `JsFormatter`.

> If you need JavaScript to implement an interface, the general API Design guidance is to create an event-based interface.  JavaScript can’t implement interfaces, but it can register for event handlers.  The other alternative is to create an “option bag” class that you would be able to construct (to defaults) and then mutate (with properties) which could then be passed into another object.  Such an object must be implemented in C++ or C++/CX.

Should we generate some type signature for interfaces or for classes to indicate they implement some interfaces?

joelmartinez:
@mpMelnikov great questions/comments

> If we decide some members can have usage sample without declaration signature, something should be done with <MemberSignature> generation.

Yes, it's clear that we're going to have to support a `MemberSignature` with a `Usage` signature, but no `Value`. You'll have to verify, firstly, that the XSD will support this configuration ... and obviously as you allude to, the code that generates the signature nodes will need to be modified accordingly to allow for this.

> For compatibility with older versions, I generate a signature of a constructor instead of class syntax from ES2015. Is it ok?

Yes that is ok

> Where can we replace IAsyncAction with a Promise in signatures?

I only used that example because it was the example given in one of the first meetings we had about this. I would recommend that you take a look at those doc links he gave for official information about what projections are appropriate.

> So, how does it affect signatures? Should we generate a signature and/or an usage sample for them? At the moment I treat generic types and methods as not supported both for declarations and samples.

Yes I _think_ this is the right answer for now. Remember that he said that the WinRT API (which is specifically where we are interested in supporting this) is designed in such a way, that any of the useful concrete types will not require generic parameters.

> Should we generate a signature of "close" method for destructors? At the moment, destructors are not supported by JsFormatter.

Yes, but only *if* the type supports `IClosable`; then I believe this would be an appropriate course of action.

> Should we generate some type signature for interfaces or for classes to indicate they implement some interfaces?

No, I think at this point we should mostly focus on the usage signatures. If there's no obvious way for the language to implement something like an interface, then I'd say to either leave it out of the signature, not generate a sig at all.

mpMelnikov:
@joelmartinez This PR seems to be ready for code review.

joelmartinez:
Ok ... I will review it. Unfortunately, much of today was taken up by some technical issues with my computer. But those have finally been resolved, so I'll finally be reviewing the c++ and this PR tomorrow morning. Thanks

mpMelnikov:
@joelmartinez We've rebased C++ PR, I've merged that branch into this one locally (there are merge conflicts which are resolved quite straightforwardly), tests work perfectly:) Could you check if it works for you?

mdoc/Test/en.expected-javascript/Mono.DocTest.Generic/GenericBase`1.xml
```diff
@@ -0,0 +1,149 @@
...
+  <Docs>
+    <typeparam name="U">To be added.</typeparam>
+    <summary>To be added.</summary>
+    <remarks>To be added.</remarks>
+  </Docs>
+  <Members>
+    <Member MemberName=".ctor">
+      <MemberSignature Language="C#" Value="public GenericBase ();" />
+      <MemberSignature Language="ILAsm" Value=".method public hidebysig specialname rtspecialname instance void .ctor() cil managed" />
+      <MemberSignature Language="JavaScript" Value="function GenericBase()" Usage="var genericBase = new GenericBase();" />
```
joelmartinez: Isn't it not possible to instantiate this class, since it requires a generic type parameter? in this case, `IsSupported` should return false for this constructor ... right?

mdoc/Test/en.expected-javascript/Mono.DocTest.Generic/GenericBase`1.xml
```diff
@@ -0,0 +1,149 @@
...
+      </ReturnValue>
+      <Docs>
+        <summary>To be added.</summary>
+        <remarks>To be added.</remarks>
+      </Docs>
+    </Member>
+    <Member MemberName="op_Explicit">
+      <MemberSignature Language="C#" Value="public static explicit operator U (Mono.DocTest.Generic.GenericBase&lt;U&gt; list);" />
+      <MemberSignature Language="ILAsm" Value=".method public static hidebysig specialname !U op_Explicit(class Mono.DocTest.Generic.GenericBase`1&lt;!U&gt; list) cil managed" />
+      <MemberSignature Language="JavaScript" Value="function op_Explicit(list)" Usage="var u = Mono.DocTest.Generic.GenericBase.op_Explicit(list)" />
```
joelmartinez: I was under the impression that conversion operators weren't supported in js ? It could very well be that `op_Explicit` can be called, but can you verify that this is in fact supported? is it mentioned anywhere in the documentation?

mdoc/Mono.Documentation/Updater/Formatters/JsUsageFormatter.cs
```diff
@@ -0,0 +1,246 @@
...
+                buf.Append(".");
+                buf.Append(GetMethodName(method));
+                buf.Append("(");
+                AppendParameters(buf, method, method.Parameters);
+                buf.Append(").done( /* Your success and error handlers */ )");
+
+                return buf.ToString();
+            }
+
+            // Non-static
```
joelmartinez: This comment is ambiguous ... it looks like it was copied in here from documentation as reference? Can you clarify in clearer prose this comment? It's not immediately clear whether this is relevant to the code before or after it.

mdoc/Test/en.expected-javascript/Mono.DocTest.Generic/GenericBase`1.xml
mpMelnikov: Yes, operators overloading is not supported by WinRT. I was a little confused by this phrase in e-mail 
>C# implementations get compiled to metadata as static methods

But it can't be compiled in WinRT project: 
`'SampleComponent.Example.op_Addition(SampleComponent.Example, SampleComponent.Example)' is an operator overload. Managed types cannot expose operator overloads in the Windows Runtime`
So, I've added this check to `IsMethodSupported` method:
```
!DocUtils.IsOperator(methodDefinition)
```

mdoc/Test/en.expected-javascript/Mono.DocTest.Generic/GenericBase`1.xml
mpMelnikov: Actually, we can't compile WinRT type if it has generic arguments or any public members with generic arguments. So, if the type isn't supported by WinRT we can treat all its members as unsupported. Will it be a good solution?

mdoc/Mono.Documentation/Updater/Formatters/JsUsageFormatter.cs
mpMelnikov: Done

mdoc/Test/en.expected-javascript/Mono.DocTest.Generic/GenericBase`1.xml
joelmartinez: Ahh, good point ... so in that case, one would never be using, say, `List<T>` directly in javascript. Yes, I think this is a fair solution

mpMelnikov:
@joelmartinez I've committed all requested changes. The commit is rebased and squashed.

joelmartinez:
🎉

PR #170, opened 2017-12-01
[mdoc] Implementation details are now shown on members

### LINKED ISSUE #158: [mdoc] Implementation details are not shown on members
We have scenarios where certain members are actually implementations of the inherited interface. As an example:

image

Here, we can see that the property implements `IBuildSettings.ApplicationUrl`. When we go to the version of [this page on docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/api/Microsoft.Build.Tasks.Deployment.Bootstrapper.BuildSettings.ApplicationUrl?view=netframework-4.7), this information is missing because it's not included in ECMAXML.

This should be part of the **M3** work.

mpMelnikov:
Node <Implements> with subnodes <ImplementsItem> show which member(s) of interface(s) is(are) implemented

Closes #158


mpMelnikov:
@joelmartinez 
Well, this task seemed a little bit easier at the first glance:)
This commit is not ready for merge yet, just wanted to get some advice whether it's an appropriate approach or not.

1. Some members can implement more than one other member, that's why I suggest this XML scheme:
```
  <Member MemberName="Method2">
      <!-- ... -->
      <Implements>
        <ImplementsItem>M:TestInterfaceImplementation.Interface5.Method2</ImplementsItem>
        <ImplementsItem>M:TestInterfaceImplementation.Interface4.Method2</ImplementsItem>
      </Implements>
      <!-- ... -->
  </Member>
```

2. Out of the box, Cecil [doesn't provide information](https://stackoverflow.com/questions/4917509/mono-cecil-something-like-type-getinterfacemap) about implemented member, that's why I've added an entity fingerprint, it's a string which is equal for members that can be an implementation of each other. So, I create a lookup table where a key is a fingerprint, and the value is a set of implemented members, and work with this table. 
Maybe there is already some similar functionality in mdoc or in Cecil? (but I haven't noticed it)

3. There are quite a lot of variants of members implementation: implicit and explicit implementation, inheritance of interfaces, generic methods in interfaces and so on. I've added a test project which contains interesting test cases. I'll include it in the integration tests.

4. Yes, I remember about XSD scheme:)

joelmartinez:
1. Yeah that makes sense ... though I'd rather use `/Implements/InterfaceMember` instead of `ImplementsItem`.
2. Interesting. yeah, up until this point, we haven't really had to do this "matching" ... because by and large that was the purview of the compiler to care about what member it was implementing. Sounds like a good solution.
3. Yes, thank you for being thorough here ... make sure you also cover cases with the `new` keyword in the member signature.
4. 😉

mpMelnikov:
@joelmartinez 
This PR is ready for code review.

mpMelnikov:
@joelmartinez I've made a commit. About the idea to replace `GetFingerprint` method with some formatter, I've considered this option and added the following remark:
```
/// <remarks>Any existing MemberFormatter can't be used for generation of fingerprint because none of them generate equal signatures
/// for an interface member and its implementation in the following cases:
/// 1. Explicitly implemented members
/// 2. Different names of generic arguments in interface and in implementing type
/// 3. Implementation of interface with generic type parameters
/// The last point is especially interesting because it makes GetFingerprint method context-sensitive:
/// it returns signatures of interface members depending on generic parameters of the implementing type (GenericInstanceType parameter).</remarks>
```
If this explanation is too vague we can discuss code examples.

joelmartinez:
@mpMelnikov thank you for the explanation ... yes it makes sense :)

mpMelnikov:
@joelmartinez oh boy. It seems we've just merged 4 commits to master:( maybe we should reset it and make a new PR?

joelmartinez:
Oh!! I didn’t even notice that. Yes, I will reset master. Please submit a new PR

joelmartinez:
@mpMelnikov ok, `master` has been reset. Please submit a new PR. 

lol, in my head, I thought these changes would have come squashed :P 

> With those changes made, you can squash the commits :)

mpMelnikov:
> I thought these changes would have come squashed :P

This time it's true:) #179 is ready

PR #175, opened 2017-12-03
mdoc: fixed breaks in frameworks mode

### LINKED ISSUE #174: mdoc 5.3.3 breaks in frameworks mode
When running the latest `mdoc` release in frameworks mode, it breaks with the following error:

```
2017-12-03T04:39:26.9611841Z ##[error]mdoc: System.ArgumentOutOfRangeException: Length cannot be less than zero.
Parameter name: length
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Mono.Documentation.Util.AttachedEntitiesHelper.GetEventName(String fieldDefinitionName)
   at Mono.Documentation.Util.AttachedEntitiesHelper.IsAttachedEvent(FieldDefinition field, Dictionary`2 methods)
   at Mono.Documentation.Util.AttachedEntitiesHelper.<GetAttachedEvents>d__8.MoveNext()
   at Mono.Documentation.Util.AttachedEntitiesHelper.<GetAttachedEntities>d__6.MoveNext()
   at Mono.Documentation.Util.CecilExtensions.<GetMembers>d__1.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at Mono.Documentation.MDocUpdater.Run(IEnumerable`1 args)
   at Mono.Documentation.MDoc.Run(String[] args)
   at Mono.Documentation.MDoc.Main(String[] args)
See `mdoc help' for more information or use --debug to diagnose.
```

Looks like the length is being reduced to below zero [here](https://github.com/mono/api-doc-tools/blob/master/mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs#L19).

mpMelnikov:
On the first phase of attached event check, we had to check name, not type
Added integration test on this case

Closes #174

### DIFF
--- a/mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs
+++ b/mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs
@@ -55,7 +55,7 @@ namespace Mono.Documentation.Util
         private static bool IsAttachedEvent(FieldDefinition field, Dictionary<string, IEnumerable<MethodDefinition>> methods)
         {
             // https://docs.microsoft.com/en-us/dotnet/framework/wpf/advanced/attached-events-overview
-            if (!field.FieldType.Name.EndsWith(EventConst))
+            if (!field.Name.EndsWith(EventConst))
                 return false;
             var addMethodName = $"Add{GetEventName(field.Name)}Handler";
             var removeMethodName = $"Remove{GetEventName(field.Name)}Handler";
--- a/mdoc/Test/AttachedEventsAndProperties/AttachedEventExample.cs
+++ b/mdoc/Test/AttachedEventsAndProperties/AttachedEventExample.cs
@@ -77,5 +77,15 @@ namespace AttachedEventsAndProperties
         {
         }
         #endregion 
+
+        #region Negative example (the event type ends with "Event", but the name doesn't)
+        public static readonly RoutedEvent NeedsCleaning7;
+        public static void AddNeedsCleaning7Handler(DependencyObject d, RoutedEventHandler handler)
+        {
+        }
+        public static void RemoveNeedsCleaning7Handler(DependencyObject d, RoutedEventHandler handler)
+        {
+        }
+        #endregion 
     }
 }
--- a/mdoc/Test/AttachedEventsAndProperties/AttachedPropertyExample.cs
+++ b/mdoc/Test/AttachedEventsAndProperties/AttachedPropertyExample.cs
@@ -39,6 +39,27 @@ namespace AttachedEventsAndProperties
             element.SetValue(IsBubbleSourceProperty, value);
         }
         #endregion
+
+        #region Negative example (the property type ends with "Event", but the name doesn't)
+
+        public static readonly DependencyProperty IsBubbleSource3 = DependencyProperty.RegisterAttached(
+  "IsBubbleSource3",
+  typeof(Boolean),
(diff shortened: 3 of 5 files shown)


mpMelnikov:
> After more investigation, looks like we might want to check if the field is of type System.Windows.RoutedEvent to make sure we are actually dealing with an attached event rather than relying on just the name.

In this code, I had to check the name but checked FieldType.Name instead.
```
//            if (!field.FieldType.Name.EndsWith(EventConst)) // old version
            if (!field.Name.EndsWith(EventConst))
                return false;
            var addMethodName = $"Add{GetEventName(field.Name)}Handler";
            var removeMethodName = $"Remove{GetEventName(field.Name)}Handler";
```

After this check, we get names of the methods which are required for attached events. And finally, we check many clues, one of them is the check of the field type, as @dend suggested.
Also, I've added 2 new test cases into `AttachedEventExample` and `AttachedPropertyExample` (`check-monodocer-attached-entities` integration test), to avoid such issues in future.

cc: @dend @joelmartinez

joelmartinez:
@mpMelnikov it seems that I merged this in prematurely 😢  After merging, I decided to do a little test, and I reverted this one line of code that does the check. Unfortunately, all of the tests (including the modification you added here) still passed ... so the test doesn't really reflect the error state. Ideally, changing line 58 back to the original code (where you're checking `FieldType`) would result in a `System.ArgumentOutOfRangeException`.

Additionally, Den also mentioned that we should be checking if the event type is a `RoutedEvent` ... is this something that should be added to this change? 

Oh, and also ... can you indicate whether you tested your PR with the assemblies Den mentioned here? https://github.com/mono/api-doc-tools/issues/174#issuecomment-348742241

This is what happens when you code-review before you've had ☕️  😉  . I've _reset_ the `master` branch to exclude this commit (not sure if you'll have to re-open a new pull request). In any case, my apologies for the confusion.

joelmartinez:
> And finally, we check many clues, one of them is the check of the field type, as @dend suggested.

Oh, I see that my "Additionally" question was already addressed ... again, apologies. My dog just woke me up and it's a bit after 6am. Seriously, don't sleep and code-review :P 

However, the other two questions stand:

- The new integration test should fail with the same error as reported in #178 when line 58 is reverted.
- The PR should be tested with the provided assemblies to ensure that the error does not occur.

mpMelnikov:
@joelmartinez I've committed a new version. Now tests fail if we revert that line of code (properties names in the previous version were too long to cause an exception). My apologies, I had to check it myself.
```
New Type: AttachedEventsAndProperties.AttachedEventExample
mdoc: Length cannot be less than zero.
Parameter name: length
See `mdoc help' for more information or use --debug to diagnose.
Makefile:445: recipe for target 'check-monodocer-attached-entities' failed
make[1]: *** [check-monodocer-attached-entities] Error 1
make[1]: Leaving directory '/data/api-doc-tools/mdoc'
Makefile:25: recipe for target 'check-mdoc' failed
make: *** [check-mdoc] Error 2

```
> The PR should be tested with the provided assemblies to ensure that the error does not occur.

Yes, I've tested it with the provided assemblies, there are no exceptions anymore.
Could you please check if it's possible to merge this pull request again or should I create a new one?

joelmartinez:
Unfortunately, it looks like you will have to open a new PR. Again, apologies for the inconvenience. Thanks for turning it t around so quickly :smile:

PR #181, opened 2017-12-08
[mdoc] Fixed StackOverflow in F# and exception in members implementations

mpMelnikov:
StackOverflow in F# was caused by generic types constrained recursively by themselves (`'T :> seq<'T>`)
Added `TestConstraints_2_2` unit test

For members implementations generation, we shouldn't count on order of type generic arguments because they can be passed to interface reordered. Slightly remade generation of fingerprints.
Extended `check-monodocer-members-implementation` integration test with `IScrollable` and `ScrollableBase` classes

### DIFF
--- a/mdoc/Mono.Documentation/Updater/DocUtils.cs
+++ b/mdoc/Mono.Documentation/Updater/DocUtils.cs
@@ -410,8 +410,8 @@ namespace Mono.Documentation.Updater
             StringBuilder buf = new StringBuilder();
             
             var unifiedTypeNames = new Dictionary<string, string>();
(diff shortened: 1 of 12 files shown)


mpMelnikov:
@joelmartinez I don't know why, but CI (Jenkins) check was not run for this pull request.

joelmartinez:
Build

joelmartinez:
build

joelmartinez:
@mpMelnikov just found out it's an infrastructure scheduled downtime of the build lab. So I will run the tests locally for the review :)

mpMelnikov:
@joelmartinez 
Oh. You're right. Now I've added them. Files in `/Test/en.*/` folders don't appear in pending changes, as they are in our `.gitignore` file, so, extra attention should be paid to them:)

PR #237, opened 2018-04-25
mdoc: Fixed tuple class incorrect representation

### LINKED ISSUE #206: Tuple class representation is incorrect
image

Note that this is incorrect syntax for F#, and the correct representation would actually be what we have on MSDN: https://msdn.microsoft.com/en-us/library/dd387185

mpMelnikov:
Added: if this is a declaration of Tuple class, treat it as an ordinary type (don't skip name)
Added a unit test for Tupple class declaration
Added skip of not public interfaces

Closes #206


mpMelnikov:
Hi @joelmartinez !

I fixed representation of `Tuple` class in F# signatures, now the only difference with MSDN is the `end` keyword. Not sure, whether we should add this keyword, or not, cause all signatures for F# are generated by `FSharpFormatter` without `end`.

PR #239, opened 2018-04-26
PR - issue 212

### LINKED ISSUE #212: [mdoc] Do not document APIs that are specially-tagged
There are cases where types are marked with `[CompilerGenerated]` and `[TypeIdentifier]`, and those should not be documented.

cc: @aarnott

katebialiatka:
check for embedded type was added - the presence of two attributes mentioned in the ticket is checked

add integration test

Closes #212

### DIFF
--- a/mdoc/Test/en.expected-embedded-type/Issue212Example/Class1.xml
+++ b/mdoc/Test/en.expected-embedded-type/Issue212Example/Class1.xml
@@ -1,0 +1,73 @@
+<Type Name="Class1" FullName="Issue212Example.Class1">
+  <TypeSignature Language="C#" Value="public class Class1 : Issue212Example.ICustomInterface" />
+  <TypeSignature Language="ILAsm" Value=".class public auto ansi beforefieldinit Class1 extends System.Object implements class Issue212Example.ICustomInterface" />
+  <AssemblyInfo>
+    <AssemblyName>DocTest-embedded-type</AssemblyName>
+    <AssemblyVersion>0.0.0.0</AssemblyVersion>
+  </AssemblyInfo>
+  <Base>
+    <BaseTypeName>System.Object</BaseTypeName>
+  </Base>
+  <Interfaces>
+    <Interface>
+      <InterfaceName>Issue212Example.ICustomInterface</InterfaceName>
+    </Interface>
+  </Interfaces>
+  <Docs>
+    <summary>To be added.</summary>
+    <remarks>To be added.</remarks>
+  </Docs>
+  <Members>
+    <Member MemberName=".ctor">
+      <MemberSignature Language="C#" Value="public Class1 ();" />
+      <MemberSignature Language="ILAsm" Value=".method public hidebysig specialname rtspecialname instance void .ctor() cil managed" />
+      <MemberType>Constructor</MemberType>
+      <AssemblyInfo>
+        <AssemblyVersion>0.0.0.0</AssemblyVersion>
+      </AssemblyInfo>
+      <Parameters />
+      <Docs>
+        <summary>To be added.</summary>
+        <remarks>To be added.</remarks>
+      </Docs>
+    </Member>
+    <Member MemberName="Count">
+      <MemberSignature Language="C#" Value="public int Count { get; }" />
+      <MemberSignature Language="ILAsm" Value=".property instance int32 Count" />
+      <MemberType>Property</MemberType>
+      <Implements>
+        <InterfaceMember>P:Issue212Example.ICustomInterface.Count</InterfaceMember>
+      </Implements>
+      <AssemblyInfo>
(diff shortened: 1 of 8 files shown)


katebialiatka:
As it was stated in the ticket description, a type is treated as embedded when it is marked with two attributes - `[CompilerGenerated]` and `[TypeIdentifier]`.

joelmartinez:
build

mdoc/Test/en.expected-embedded-type/Issue212Example/Class1.xml
```diff
@@ -0,0 +1,73 @@
...
+  <AssemblyInfo>
+    <AssemblyName>DocTest-embedded-type</AssemblyName>
+    <AssemblyVersion>0.0.0.0</AssemblyVersion>
+  </AssemblyInfo>
+  <Base>
+    <BaseTypeName>System.Object</BaseTypeName>
+  </Base>
+  <Interfaces>
+    <Interface>
+      <InterfaceName>Issue212Example.ICustomInterface</InterfaceName>
```
joelmartinez: @dend could there be any downstream issues, with the fact that this is now referencing an interface that _does not exist_ in the EcmaXML? For example, we currently [link to interfaces implemented by a class](https://docs.microsoft.com/en-us/dotnet/api/system.string?view=netframework-4.7.1)

image

In this case, that would lead to a 404. Does there need to be a corresponding task to this, to hide the link on the front-end for interfaces that "don't exist" (from the perspective of EcmaXML)?

joelmartinez: To be clear, this **may** only be an issue with this specific and contrived test case we have here ... in "real life", I'm not sure under what conditions we'll encounter an embedded interface like this. Will the name exactly match the same named type in another assembly that we'll definitely have onboarded (in which case this is a non-issue)?

joelmartinez: Perhaps @AArnott or @terrajobst might have some insight here as well :)

dend: @joelmartinez correct me if I am wrong, but if we don't document an interface that is embedded, it will not be available _anywhere_ in ECMAXML because documentation for it is not generated? Or do you mean that say, `Class_A : ISecret` will result in us showing that it still implements `ISecret`?

joelmartinez: @dend I was going off of @terrajobst's comment here: https://github.com/mono/api-doc-tools/issues/212#issuecomment-372421325

> Of course, we should still document **the original definitions (PIA)** but not the no-PIA embedded copies. And that's what the attributes help with.

and @AArnott's comment here: https://github.com/mono/api-doc-tools/issues/212#issuecomment-371345947

> Such interfaces should only be documented **where they are originally defined: in the embeddable assembly**. Any time the same type is found as an embedded type in another assembly (identifiable by the presence of the two attributes mentioned in the issue description), this should not be a clue that the docs.microsoft.com page should claim the interface is also available from that assembly.

So this suggests to me that these interfaces _will_ exist somewhere

joelmartinez: @dend also

> if we don't document an interface that is embedded, it will not be available anywhere in ECMAXML because documentation for it is not generated

Correct, as implemented right now, that interface will not be in *this* set EcmaXML files (thus the question above).

> Or do you mean that say, Class_A : ISecret will result in us showing that it still implements ISecret?

Also correct ... we're blocking the documenting of "`ISecret`", but there are still traces of it, like in the interfaces, and in the signature, etc

dend: I _believe_ we can just exclude those entirely in the view with the help of our filtering mechanism. @joelmartinez

joelmartinez: Ahh, yeah good call! So does that mean you're 👍 on this? Who would be responsible for coming up with the list of types to be excluded by the filter?

dend: Content teams 😀 Once this makes it into production, we can assess how it impacts existing content. @joelmartinez

AArnott: We have many VS SDK assemblies that are embeddable. And *most* are publicly available and documented. Not all are, and those that aren't, if the interface is either not displayed as implemented by a class in documentation, or it is displayed but no link is offered to jump to the non-existent documentation, either way sounds fine IMO, since the interface in that case is an implementation detail.

Issue #227, opened 2018-04-12
monodoc.dll.config does not substitute monodoc_refdir


Therzok:
https://github.com/mono/api-doc-tools/blob/c503071997d1cfcdb5a039a8a36e530412cb0acb/monodoc/monodoc.dll.config#L3

This is not substituted in the shipped nuget package

sandyarmstrong:
An option on typical Mac systems is to use `RootTree.LoadTree("/Library/Frameworks/Mono.framework/Versions/Current/lib/monodoc/");`

PR #246, opened 2018-05-02
Add a MonoDoc nuget package

### LINKED ISSUE #228: [NuGet] Ship monodoc.dll in build directory
monodoc.dll can be referenced by applications to integrate with mono documentation.

mhutch:
Fixes #228

### DIFF
--- a/monodoc/Test/Monodoc.Test.csproj
+++ b/monodoc/Test/Monodoc.Test.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net461</TargetFramework>
+    <TargetFramework>net471</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
     <!-- PrivateAssets=all is to keep VSMac happy by copying NUnit.Framework.dll -->
--- a/monodoc/Makefile
+++ b/monodoc/Makefile
@@ -1,5 +1,5 @@
 CONFIGURATION = Release
 
 check: 
-	mono ../packages/NUnit.ConsoleRunner.3.8.0/tools/nunit3-console.exe Test/bin/$(CONFIGURATION)/Monodoc.Test.dll
+	mono ../packages/NUnit.ConsoleRunner.3.8.0/tools/nunit3-console.exe Test/bin/$(CONFIGURATION)/net471/Monodoc.Test.dll
 	@echo "monodoc Tests Complete!"


Therzok:
Does this fix https://github.com/mono/api-doc-tools/issues/227 ?

Otherwise, it won't work

mhutch:
No, it doesn't

Makefile
```diff
@@ -8,12 +8,11 @@ all: build
 build: $(MDOC)
 
 $(MDOC):
-	$(MSBUILD) apidoctools.sln /p:Configuration=$(CONFIGURATION);
+	$(MSBUILD) apidoctools.sln /r /p:Configuration=$(CONFIGURATION);
```
mhutch: This does an MSBuild NuGet restore

Makefile
```diff
@@ -26,6 +25,7 @@ check-mdoc:
 
 nuget:
 	nuget pack mdoc/mdoc.nuspec -outputdirectory bin/Nuget
+	$(MSBUILD) monodoc/monodoc.csproj /p:Configuration=$(CONFIGURATION) /t:Pack
```
mhutch: This runs the MSBuild pack target

Package.props
```diff
@@ -0,0 +1,11 @@
+<Project>
```
mhutch: General package metadata

apidoctools.sln
```diff
@@ -9,8 +9,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "monodoc", "monodoc\monodoc.
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Monodoc.Test", "monodoc\Test\Monodoc.Test.csproj", "{1EE70E2C-A289-4C36-AD0A-3D0C6CE56615}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ICSharpCode.SharpZipLib", "external\SharpZipLib\ICSharpCode.SharpZipLib.NET45\ICSharpCode.SharpZipLib.csproj", "{0E7413FF-EB9E-4714-ACF2-BE3A6A7B2FFD}"
```
mhutch: Switched this to a NuGet reference so it could be a package dependency

mdoc/mdoc.Test/CppCxFormatterMembersTests.cs
```diff
@@ -126,7 +126,7 @@ public void NoSupport_DefaultParameters()
             TestMethodSignature(CSharpTestLib, "Mono.DocTest.Widget", "Default", null);
         }
         
-        [TestFixtureTearDown]
+        [OneTimeTearDown]
```
mhutch: Tracking NUnit API change

mdoc/mdoc.Test/mdoc.Test.FSharp/mdoc.Test.FSharp.fsproj
```diff
@@ -94,16 +94,12 @@
...
-      <HintPath>..\..\..\packages\FSharp.Core.4.3.4\lib\net45\FSharp.Core.dll</HintPath>
-    </Reference>
     <Reference Include="mscorlib" />
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Numerics" />
-    <Reference Include="System.ValueTuple">
-      <HintPath>..\packages\System.ValueTuple.4.3.1\lib\netstandard1.0\System.ValueTuple.dll</HintPath>
-    </Reference>
+    <PackageReference Include="FSharp.Core" Version="4.3.4" />
```
mhutch: Switched these to PackageReferences for consistency

monodoc/Test/Monodoc.Test.csproj
```diff
@@ -1,59 +1,11 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
mhutch: SDK style projects are much nicer to hand edit :)

monodoc/monodoc.csproj
```diff
@@ -1,713 +1,27 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
mhutch: Had to switch to SDK style to get netstandard 2.0 support and the MSBuild Pack target.

monodoc/monodoc.csproj
```diff
@@ -1,713 +1,27 @@
...
-    <DefineConstants>DEBUG;LEGACY_MODE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <ConsolePause>false</ConsolePause>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <Optimize>true</Optimize>
-    <OutputPath>..\bin\Release</OutputPath>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
mhutch: Yay netstandard 2.0!

monodoc/monodoc.csproj
mhutch: Also allows MASSIVELY simplifying the project file

monodoc/monodoc.csproj
```diff
@@ -1,713 +1,27 @@
...
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources/images/treebg.png">
-      <LogicalName>treebg.png</LogicalName>
-    </EmbeddedResource>
-  </ItemGroup>
-  <ItemGroup>
-    <Content Include="monodoc.dll.config">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </Content>
+    <PackageReference Include="SharpZipLib" Version="1.0.0-alpha2" />
```
mhutch: One of the nice things about PackageReference is that the Pack target automatically converts it into a NuGet package dependency.

monodoc/monodoc.csproj
```diff
@@ -1,713 +1,27 @@
...
+    <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.1" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\external\SharpZipLib\ICSharpCode.SharpZipLib.NET45\ICSharpCode.SharpZipLib.csproj">
-      <Project>{0e7413ff-eb9e-4714-acf2-be3a6a7b2ffd}</Project>
-      <Name>ICSharpCode.SharpZipLib</Name>
-    </ProjectReference>
+    <Compile Include="..\external\Lucene.Net.Light\src\core\**\*.cs" Link="Lucene\%(Filename)%(Extension)" />
+    <Compile Remove="Mono.Utilities\MemoryLRU.cs" />
```
mhutch: These files needed to be explicitly removed from the wildcard to match the old project's explicit file list

monodoc/Monodoc/generators/html/Ecmaspec2Html.cs
```diff
@@ -10,7 +10,7 @@ namespace Monodoc.Generators.Html
 	public class Ecmaspec2Html : IHtmlExporter
 	{
 		static string css_ecmaspec;
-		static XslTransform ecma_transform;
+		static XslCompiledTransform ecma_transform;
```
mhutch: Updating obsoleted API

mhutch:
@joelmartinez I added a bunch of comments :)

mhutch:
Currently failing with

```
 /usr/local/share/dotnet/sdk/2.1.200/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(198,5): warning : A stable release of a package should not have a prerelease dependency. Either modify the version spec of dependency "SharpZipLib [1.0.0-alpha2, )" or update the version field in the nuspec. [/Users/builder/jenkins/workspace/Docs-MonoApiDocTools/monodoc/monodoc.csproj]
```

joelmartinez:
Ok, I took a look at this, and it seems like the newer version of nunit changed something about the way paths are resolved (maybe the default working directory, or something) ... so as a result, a bunch of tests that have to load some assemblies are failing.

I played around and got some of them passing by making this one change:

```
diff --git a/mdoc/mdoc.Test/BasicTests.cs b/mdoc/mdoc.Test/BasicTests.cs
index cddbcbd..ef136aa 100644
--- a/mdoc/mdoc.Test/BasicTests.cs
+++ b/mdoc/mdoc.Test/BasicTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.IO;
 using Mono.Cecil;
 
 namespace mdoc.Test
@@ -18,7 +19,10 @@ namespace mdoc.Test
 
             if (!moduleCash.ContainsKey(filepath))
             {
-                var readModule = ModuleDefinition.ReadModule(filepath);
+                var testAssemblyFilename = this.GetType ().Assembly.Modules.First().FullyQualifiedName;
+                var testDirectory = Path.GetDirectoryName (testAssemblyFilename);
+                var assemblyToOpenPath = Path.Combine (testDirectory, filepath);
+                var readModule = ModuleDefinition.ReadModule(assemblyToOpenPath);
                 moduleCash.Add(filepath, readModule);
             }
```
But there are still some other tests failing ... I've got to focus on putting a release together, so I'm probably not going to be able to look at this for another two weeks or so (given upcoming travel, etc). 

Is there a way that maybe the scope of changes could be scaled back a bit? I fully understand and agree that SDK style projects are much nicer, but maybe that should be a standalone change/upgrade from the monodoc nuget? same for some of the other changes in this PR ...

mhutch:
@joelmartinez the reason I migrated to SDK style projects was specifically because it makes it much easier to define/build the nuget packages though. I can't really separate them.

joelmartinez:
@mhutch ok, fair enough ... then could it perhaps be decoupled from upgrading nunit (assuming that was the change that otherwise broke the unit tests)?

mhutch:
@joelmartinez unfortunately nunit 2.6.4 doesn't support netstandard... :(

joelmartinez:
@mhutch gotcha ... ok, I'll try to keep working on getting the unit tests passing here, but as I mentioned above, this will probably land after I'm able to get 5.7 out (ie. in a few weeks) :)

mhutch:
fair enough!

kekyo:
Ping: I'm waiting packaged monodoc :)  what suspended reason?

joelmartinez:
@kekyo It is on my agenda to revisit this soon ... no specific ETA, but it's definitely on the agenda. The PR needs some changes to ensure the unit tests still pass, not to mention some recent `master` changes that will have to be carefully merged

joelmartinez:
I'm going to be breaking out some of these changes into separate/individual commits ... starting with updating to the latest nunit. This way I can make sure that all the unit tests continue passing as we subsequently move to SDK style projects, etc

PR #250, opened 2018-05-07
Allow `mdoc assemble` to assemble docs from a single framework

### LINKED ISSUE #194: Allow `mdoc assemble` to assemble docs from a single framework
when using frameworks mode, sometimes it would be useful to `mdoc assemble`, and have the artifacts represent a single framework.

mpMelnikov:
Interim commit.
Added fx/framework CLI parameter for the assemle command
Filter out types, which are not mentioned in the framework file

Closes #194


mpMelnikov:
Hi @joelmartinez!

It's a temporary commit. Here, we filter out types, which are not mentioned in a framework index file. Could you please take a look at the changes, and tell if we are moving in right direction.
There are going to be added tests and filtering of members with `@FrameworkAlternate` attribute.

joelmartinez:
@mpMelnikov so far from a casual glance at the code, it's looking good! a few notes:

- I notice a lot of whitespace-only changes ... try to minimize those as it will make code review simpler.
- I'm interested in seeing tests ... I don't _think_ we have any existing `assemble` tests (do we?). But it might be difficult because the end result is a binary file (and a .tree file). That binary file is just a zip, so technically speaking you could unzip it and then do tests there? I don't know if that would be too onerous ... am interested in hearing thoughts/ideas on how best to test this. Perhaps you could write some extensive unit tests on the parts you add, to make sure that the nodes are being filtered out, and that might be enough?
- This is starting to add some reading of the framework index files. I wonder if we should have you move these new classes that deal with it *out* of the assembler namespace, and into a more general/shareable namespace ... because I'm thinking this might be useful in other areas of the code, for example generating intellisense based on a framework.

mpMelnikov:
> I notice a lot of whitespace-only changes 

Oh... Sorry for the inconvenience, will fix it when squash the commits, and will try to avoid it in the future.

> This is starting to add some reading of the framework index files. I wonder if we should have you move these new classes that deal with it out of the assembler namespace

Done! Added them to `Mono.Documentation.Framework` namespace

> I'm interested in seeing tests ...

> Perhaps you could write some extensive unit tests on the parts you add, to make sure that the nodes are being filtered out, and that might be enough?

I've implemeted this variant. Unit tests are faster, it's easy to find the problem if a unit test is failed, and there is no technical issues to run them:) The new unit tests cover new functionality. The integration test could cover reading from discs and other logic, which is not covered by these unit tests. Well, I don't now if we need it. What do you think?

joelmartinez:
> ... sorry ...

No need to apologize ... actually it's more my fault than anything. Back before you guys first started working, I knew that we were going to be doing a lot of work on the updater, so I reformatted the whole source file using modern/default visual studio settings (https://github.com/mono/api-doc-tools/commit/8822f7729c04ba22acbe9a9f1141d181ed725b6a) in a separate commit. That way, when you started working, this wouldn't be an issue. But I neglected to do that here before you started working on this file. And for that **I** apologize :P

> unit tests

👍 I'm cool with this ... one comment on the tests, please make sure you include tests where *no framework* is passed in. mdoc still has to support docsets that do not use "frameworks mode", and as such they won't be passing anything in for the fx name. 

I'm also cool with not doing any integration tests, as that will likely be quite technically difficult due to the nature of the file artifacts. **though**, I will request that you just manually do some before submitting the PR for final review ... just as a sanity check. Basically, I'm going to do the same thing, but at least this way you can catch any obvious issues with both frameworks mode `assemble`, and non frameworks mode.

Thanks!

mpMelnikov:
@joelmartinez 

I've added the tests, tested framework-mode manually and squashed the commits.

mpMelnikov:
Hi @joelmartinez!
Is it time for me to rebase this commit on master? Or is there something else to change?

joelmartinez:
@mpMelnikov hi ... yes, that would be helpful to make sure it's easily mergeable. However, I haven't really reviewed this one quite yet. I'm going to try to do the release of 5.7 _without_ this particular feature, as it's not a critical component for milestone#3. As soon as I get 5.7 out, I will return to this and review it ... as I'll need it for an upcoming project myself :P

PR #255, opened 2018-05-16
Fix for EII properties names

### LINKED ISSUE #92: Explicit Interface Implementation properties and indexers have wrong name
We found that if a member is EII (Explicit Interface Implementation), and if it's a Property or Indexer, its name will have an extra interface name in it, which is wrong. For example class `Microsoft.VisualBasic.Collection` has an member that implements `System.Collections.IList.IsFixedSize`, however in mdoc generated ECMAXML its name is:
`<Member MemberName="System.Collections.IList.IListIsFixedSize">`, note **IList**IsFixedSize has extra **IList** prefix.

katebialiatka:
Closes #92


katebialiatka:
@joelmartinez Please, find below some comments on points you have mentioned in the issue.

1) > Not sure if this is necessary, but consider adding a new value to the member node to track the ExplicitInterfaceName (if it differs from the member name we get from the metadata reflection). If you decide to add this, make sure you modify the XSD accordingly.

There was no need to bring up new nodes - I've just altered the way how existing code is used to fix the issue. [Here](https://github.com/mono/api-doc-tools/pull/255/commits/ebe3326bd94d758fa611b02a39f55b5654246a0e#diff-f0573f2c138c1eefbc878f13cb31384dR190) instead of VB property name value from the interface is passed.

2) > We have to make sure that the docids that we generate, match the docIDs that both mono.cecil, and the compiler's /doc flag generates.
Next steps were done:
- to check mono.cecil docIDs 
Command `DocCommentId.GetDocCommentId`
Result `P:CustomNamespace.ClassEnumerator.CustomProp2`
- to check compiler's /doc flag 
The appropriate flag was set to the properties of the project
Result `CustomNamespace.ClassEnumerator.CustomProp1`
- to check our docids 
Command `slashdocFormatter.GetName(<MemberReference for Eii property>)`
Result `CustomNamespace.ClassEnumerator.CustomNamespace#CustomInterface#CustomProp2`

As you can see, the last point shows some strange result. Could you please doublecheck whether we used appropriate commands to check how the last item is generated? And what is the desired result for the last point?

3)  > In a previous implementation, I went to great lengths to work around this issue by using some naming conventions commonly used in some BCL VB libraries ... this code will probably have to be reviewed, and changed or removed.

It looks like the fix which we applied for EII properties cannot cover all cases. So not to break existing functionality, it was decided to leave as is. If you have any ideas how to check this code after refactoring to cover all cases, then we can discuss what is possible to do here =)

4) > The SlashDocMemberFormatter is used to generate what is essentially docIDs. You should verify that any changes you make will not cause any regressions in how this formatter is used.

I left is without changes. Because of this point now there are two separate methods  [in the DocUtils class](https://github.com/mono/api-doc-tools/pull/255/commits/ebe3326bd94d758fa611b02a39f55b5654246a0e#diff-f0573f2c138c1eefbc878f13cb31384dR133)  - old one `GetMember` for SlashDocMemberFormatter and second one GetMemberForProperty for this ticket.

5) > You will need to make sure that the DocumentationEnumerator and EcmaDocumentationEnumerator classes are equipped to handle this change in how we're naming things ... it needs to be able to match the reflected member, with the XML node, given its name.

Changes are done in DocumentationEnumerator only but are applied to EcmaDocumentationEnumerator too because of inheritance.

6) > Related to the prior point, this code needs to be able to migrate existing EcmaXML repositories, that may or may not be using the "old" name ... because if it doesn't match the existing nodes to the right member, it will probably end up deleting the existing node (along with documentation), and just making a new/blank one.

I've added [an integration test](https://github.com/mono/api-doc-tools/pull/255/commits/ebe3326bd94d758fa611b02a39f55b5654246a0e#diff-624032a038198b3b4668cd40783d66a9R198) 
to check whether import will work correclty - ECMA XML has VB names and output XML has interface names with appropriate doc node from ECMA XML

joelmartinez:
@katebialiatka 

1. But the issue here is that the EcmaXML is _losing_ some information ... while the `@Name` property should follow the same convention that a C# assembly does (using `Prop2` from the test example). But I'd _like_ to keep the custom name in the metadata (ie. `CustomProp2`).
2. Right, so this is what I was afraid of ... this should match the other two so that we can properly import the triple slash documentation using `-i`
3. As long as _1)_ it's consistent, and _2)_, existing docs can still be successfully imported (for example, bcl)
4. ... no comment here, except that you should make sure that `#2` is worked out.
5. 👍 
6. 👍

katebialiatka:
@joelmartinez I've looked at the issue with 2) point.  It appeared that -doc option works in a different way for different programming languages (keeping in ming Eii properties):
- C# compiler generates docIds with `#`. Eg, `P:Mono.DocTest.Generic.MyList`2.System#Collections#Generic#ICollection{A}#IsReadOnly`
- for other languages (Cpp, VB, F# were checked), the compiler generates a simple signature. Eg, `P:Mono_DocTest_Generic.MyList1`2.IsReadOnly`

Because of this difference, there was a need to identify, which language compiler was used to generate triple slash documentation. XML with documentation does not contain any flag to identify, which language to choose. So, to fix this I come up with next solution - have 2 types of formatters: `SlashDocCSharpMemberFormatter `- to generate C# compiler docIDs (which uses `#` as a delimiter), and `SlashDocMemberFormatter` - to generate signatures for other compilers.

[Integration test](https://github.com/mono/api-doc-tools/pull/255/commits/522acf9f923524afa1a7c4df743413991cea9fdb#diff-624032a038198b3b4668cd40783d66a9R204) for slash doc import for VB was added. For C# slash doc import, unit tests are already present.

katebialiatka:
@joelmartinez Also please find below several considerations about this issue:

7) Interface property name was added to all formatters (including `MemberName` attribute), except ILMember formatter - in the same manner as for methods. It looks reasonable as IL needs to represent code as is. So `Member` node will [look like this](https://github.com/mono/api-doc-tools/pull/255/commits/522acf9f923524afa1a7c4df743413991cea9fdb#diff-675fbe466e074fcc3a40c2a131b532adR45).

8) During fixing the ticket, there was an issue found with c++ formatters - Eii property names has invalid delimiter in its signature. This was fixed with adding delimiter as parameter [to the method.](https://github.com/mono/api-doc-tools/pull/255/commits/522acf9f923524afa1a7c4df743413991cea9fdb#diff-f0573f2c138c1eefbc878f13cb31384dR175) Unit test was added to cover this. 

9) Also, the similar issue is with F# usage formatter. However, this needs more attention and investigations.
Currently, F# usage signature for EII interface is generated as:
`<MemberSignature Language="F#" Value="" Usage="CustomNamespace.CustomInterface.Prop1" />`
However, [F# article says](http://firstclassthoughts.co.uk/Articles/Programming/ExplicitAndImplicitInterfaceImplementationInFSharp.html#explicit-interface-implementation-in-fsharp) that it should look like:
`let luckyLuke = (new Cowboy() :> ICowboy)                   
                            luckyLuke.Draw()`
Do I need to create a ticket to investigate the problem why our signature differs from one from the article?

10) I was able to add build of VB code to the integration tests:
- `ClassEnumerator.vb` file was added as the source code
-  script to run unit tests needs to install `mono-vbnc`; also installation of `tzdata` was added to this script to fix issues with local time on the latest Ubuntu image for Docker.

@joelmartinez Could you please review the code and share your thoughts?

mdoc/Test/en.expected-eii-implementation/CustomNamespace/ClassEnumerator.xml
```diff
@@ -0,0 +1,116 @@
...
+      <ReturnValue>
+        <ReturnType>System.Int32</ReturnType>
+      </ReturnValue>
+      <Docs>
+        <summary>To be added.</summary>
+        <value>To be added.</value>
+        <remarks>To be added.</remarks>
+      </Docs>
+    </Member>
+    <Member MemberName="CustomNamespace.CustomInterface.Prop2">
```
joelmartinez: I still think we need to have a new attribute here in order to capture the fact that in the code, it's actually called `CustomProp2` ... there should be a way to detect when these do not match in the case of EII, and add something like `ExplicitInterfaceMemberName="CustomProp2"`, or something.

Think about it from a customer perspective ... if they need information about this member, they will look for `CustomProp2`, so this needs to be a part of the structured data, so that it gets indexed by the search engine accordingly.

mdoc/Consts.cs
```diff
@@ -22,7 +22,9 @@ public static class Consts
 
 		public const string DependencyPropertyFullName = "System.Windows.DependencyProperty";
 		public const string DependencyObjectFullName = "System.Windows.DependencyObject";
-		public const string VoidFullName = "System.Void";
+		public const string VoidFullName = "System.Void";
 		public const string RefTypeObsoleteString = "Types with embedded references are not supported in this version of your compiler.";
+		public const string CompilerGeneratedAttribute = "System.Runtime.CompilerServices.CompilerGeneratedAttribute";
```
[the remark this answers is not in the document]
katebialiatka: Fixed automatically after rebase

katebialiatka:
Please, find answers below.

> if I'm understanding correctly that you're using both SlashDocCSharpMemberFormatter and SlashDocMemberFormatter when importing, so that if you don't find the member using one, you'll use the other?

Yes, you are completely right! 

> re: 9., good catch ... yes, this is correct, you have to explicitly cast to the interface in order to use EII members. But actually, this is the case with C# as well isn't it? You can/should probably just make a new GH issue for this and we can address it in a future update as it's kind of a separate concern.

The issue was created - see [#264](https://github.com/mono/api-doc-tools/issues/264)

> I want to make sure that we're not duplicating work here, please check this other PR to make sure that the MsxdocSlashDocMemberFormatter isn't the same as the SlashDocCSharpMemberFormatter you've implemented here ... can you detail the differences in these two and verify that they won't conflict when merged into master?

So, they have a different purpose:
`SlashDocCSharpMemberFormatter` - it is an old version of `SlashDocMemberFormatter`; it is used to map triple slash docs generated by C# compiler. And currently,  `SlashDocMemberFormatter` contains new code to map triple slash docs generated by C++, F# etc compilers (as they use quite a different syntax for DocIds)
`MsxdocSlashDocMemberFormatter` - was added in another PR to fix issues with returning ref types.

Both functionalities need to be preserved in the mdoc. However, probably it will need some refactoring after merging to master. To overcome this, I would like to suggest you to merge this both PRs one by one (order does not matter) - when one of them is merged, we'll do rebase, solve conflicts, refactor code after that, test it a bit and then it will be possible to merge the second PR to master without any pain.

mdoc/Test/ClassEnumeratorECMA.xml
```diff
@@ -0,0 +1,125 @@
+<!-- ====================================================================== -->
```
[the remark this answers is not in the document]
katebialiatka: Frankly speaking, I created this file manually in the same manner as it was done for `TestEcmaDocs.xml`. So this comment was preserved from the file mentioned.
Do I need to clean it up?

mdoc/Test/en.expected-eii-implementation/CustomNamespace/ClassEnumerator.xml
katebialiatka: Done.
Attribute `ExplicitInterfaceMemberName` is added each time when Eii implementation name differs from the original interface name.

A fix is applied in two places in the code to be sure that all cases are covered (initial doc generation and when the update is done).

katebialiatka:
@joelmartinez  Fix for the new attribute is added.
Attribute `ExplicitInterfaceMemberName` is added each time when Eii implementation name differs from the original interface name.

Could you please review changes and approve them if needs?

joelmartinez:
@katebialiatka ok this looks great! **one last thing** ... can you please amend the .xsd to include the new `ExplicitInterfaceMemberName` attribute?

katebialiatka:
@joelmartinez  Done!
Checked with `validate` command on the newly generated file.

joelmartinez:
build

joelmartinez:
Hmm, I tried to do a best-guess merge ... but it seems that the newly-merged branch results is failing the integration tests :/ 

@katebialiatka can you take a look at this?

mpMelnikov:
Hi @joelmartinez!
Kate is on vacation now. I'll try to deal with this merge tomorrow.

joelmartinez:
@mpMelnikov much appreciated! Though, not sure if you usually work on saturdays ... but if not, please feel free to wait until next monday. I'm putting together a build today, and I'm just going to move this particular item/issue into the next release which will come later 😄

mpMelnikov:
I think it is fixed, there are little changes: now the importer uses correct signatures to find fitting members in XML.

But it was a hard nut to crack to make integration tests run successfully:) I use Docker to run them, but my Ubuntu-based image couldn't build the project as it had been upgraded to .NET Framework 4.7.1. I updated `mono-devel` to the latest version, but it failed integration tests, mdoc generates different lists of exceptions in the tests with `--exceptions=all` flag (it's 41 tests). It looks like such tests really depend on environment. 

The workaround to run the tests locally was downgrading the project to .NET Framework 4.5 and replacing Tuples with KeyValuePair (I did not commit these changes). I hope CI will complete the tests successfully. In future, maybe it makes sense to add some docker config, so everybody could run tests in the same environment?

joelmartinez:
build

mdoc/Test/en.expected-eii-implementation-ecmadoc/CustomNamespace/ClassEnumerator.xml
```diff
@@ -0,0 +1,116 @@
...
+        <remarks>ECMA ctor remarks</remarks>
+      </Docs>
+    </Member>
+    <Member MemberName="CustomNamespace.CustomInterface.Prop1" ExplicitInterfaceMemberName="CustomProp1">
+      <MemberSignature Language="C#" Value="int CustomNamespace.CustomInterface.Prop1 { get; set; }" />
+      <MemberSignature Language="C++ CLI" Value="property int CustomNamespace::CustomInterface::Prop1 { int get(); void set(int value); };" />
+      <MemberSignature Language="C++ CX" Value="property int CustomNamespace::CustomInterface::Prop1 { int get(); void set(int value); };" />
+      <MemberSignature Language="F#" Value="" Usage="CustomNamespace.CustomInterface.Prop1" />
+      <MemberSignature Language="ILAsm" Value=".property instance int32 CustomProp1" />
+      <MemberSignature Language="VB.NET" Value=" Property Prop1 As Integer Implements CustomInterface.Prop1" />
```
joelmartinez: Shouldn't the VB.NET signature here be `Property CustomProp1 As Integer Implements CustomInterface.Prop1`? It shows `ExplicitInterfaceMemberName="CustomProp1"`

mdoc/Test/en.expected-eii-implementation-ecmadoc/CustomNamespace/ClassEnumerator.xml
```diff
@@ -0,0 +1,116 @@
...
+        <remarks>ECMA CustomProp1 remarks</remarks>
+      </Docs>
+    </Member>
+    <Member MemberName="CustomNamespace.CustomInterface.Prop2" ExplicitInterfaceMemberName="CustomProp2">
+      <MemberSignature Language="C#" Value="object CustomNamespace.CustomInterface.Prop2 { get; }" />
+      <MemberSignature Language="C++ CLI" Value="property System::Object ^ CustomNamespace::CustomInterface::Prop2 { System::Object ^ get(); };" />
+      <MemberSignature Language="C++ CX" Value="property Platform::Object ^ CustomNamespace::CustomInterface::Prop2 { Platform::Object ^ get(); };" />
+      <MemberSignature Language="F#" Value="" Usage="CustomNamespace.CustomInterface.Prop2" />
+      <MemberSignature Language="ILAsm" Value=".property instance object CustomProp2" />
+      <MemberSignature Language="VB.NET" Value=" ReadOnly Property Prop2 As Object Implements CustomInterface.Prop2" />
```
joelmartinez: Same question here ... looks like the VB.NET signature should be using `CustomProp2`

mpMelnikov:
@joelmartinez, the little fix is made, now VB signatures for EII properties are generated based on the name, which is specified in VB source code:

```csharp
- buf.Append(DocUtils.GetPropertyName(property, NestedTypeSeparator).Split('.').Last());
+ buf.Append(property.Name.Split('.').Last());
```

joelmartinez:
Awesome, thank you so much for getting this tweaked! 👏

Issue #293, opened 2018-07-18
NullReferenceException when reflecting binaries


dend:
See [recent build](https://github.com/mono/api-doc-tools/releases/download/tag-mdoc-5.7.2/mdoc-5.7.2.zip).

```
2018-07-18T17:40:09.7953866Z Updating: System.ParamArrayAttribute
2018-07-18T17:40:09.7979485Z Updating: System.PlatformNotSupportedException
2018-07-18T17:40:09.8014852Z Updating: System.Predicate`1
2018-07-18T17:40:09.8036866Z Updating: System.RankException
2018-07-18T17:40:09.8140995Z Updating: System.ReadOnlyMemory`1
2018-07-18T17:40:09.8234998Z Updating: System.ReadOnlySpan`1
2018-07-18T17:40:11.0188471Z ##[error]mdoc: System.NullReferenceException: Object reference not set to an instance of an object.
   at Mono.Documentation.MDocUpdater.IsReadonlyAttribute(IList`1 attributes)
   at Mono.Documentation.MDocUpdater.MakeReturnValue(FrameworkTypeEntry typeEntry, XmlElement root, TypeReference type, IList`1 attributes, Boolean shouldDuplicateWithNew)
   at Mono.Documentation.MDocUpdater.MakeReturnValue(FrameworkTypeEntry typeEntry, XmlElement root, MemberReference mi, Boolean shouldDuplicateWithNew)
   at Mono.Documentation.MDocUpdater.UpdateMember(DocsNodeInfo info, FrameworkTypeEntry typeEntry, Dictionary`2 implementedMembers)
   at Mono.Documentation.MDocUpdater.DoUpdateType2(String message, XmlDocument basefile, TypeDefinition type, FrameworkTypeEntry typeEntry, String output, Boolean insertSince)
   at Mono.Documentation.MDocUpdater.DoUpdateType(AssemblySet set, TypeDefinition type, FrameworkTypeEntry typeEntry, String basepath, String dest)
   at Mono.Documentation.MDocUpdater.DoUpdateAssembly(AssemblySet assemblySet, AssemblyDefinition assembly, XmlElement index_types, String source, String dest, HashSet`1 goodfiles)
   at Mono.Documentation.MDocUpdater.DoUpdateAssemblies(String source, String dest)
   at Mono.Documentation.MDocUpdater.Run(IEnumerable`1 args)
   at Mono.Documentation.MDoc.Main(String[] args)
[...]

PR #294, opened 2018-07-20
mdoc 5.7.2.1

joelmartinez:
- #92 - Explicitly Implemented Members that use a custom name (VB) are now named using the same naming convention as C# EII members. There is an additional `ExplicitInterfaceMemberName` attribute on the `Member` node in these cases that have the custom name.
- #293 - fixes NRE.
- Additional minor perf improvement with assembly resolution.


joelmartinez:
build

PR #307, opened 2018-08-17
[WIP] Modernizing the .csproj files

mattleibow:


joelmartinez (changes requested):
@mattleibow looks like the test failure is actually this:

```
mono ../bin/Release/mdoc.exe update -o Test/en.actual Test/DocTest-VB-Eii.dll -lang vb.net -lang fsharp -lang javascript -lang c++/cli -lang c++/cx -lang c++/winrt -import Test/ClassEnumeratorSlashDoc.xml
Cannot open assembly '../bin/Release/mdoc.exe': No such file or directory.
```
Looks like the csproj is not outputting the project to the expected location (`bin/Release/mdoc.exe` ... in the root of the repo). This is where the integration tests expect the build artifact to be.

Once you update that, all the tests should pass :)

Issue #319, opened 2018-08-21
F#: list members for classes


cartermp:
Currently, no members are listed for F# classes. However, in two places this is done:

* Tooltips in all F# editors
* The existing MSDN reference for FSharp.Core ([example](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/control.async-class-%5bfsharp%5d))

So I think they should be listed in the signature. F# developers have been trained to read about the members a type has in the signature like this.

Issue #318, opened 2018-08-21
F#: structs should use the Struct attribute in the signature


cartermp:
Example:
![image](https://user-images.githubusercontent.com/6309070/44372954-64090900-a49b-11e8-8147-cb133a162823.png)

This type is actually declared in source as:

```fsharp
    [<StructuralEquality; StructuralComparison>]
    [<CompiledName("FSharpResult`2")>]
    [<Struct>]
    type Result<'T,'TError> = 

      /// Represents an OK or a Successful result. The code succeeded with a value of 'T.
      | Ok of ResultValue:'T 

      /// Represents an Error or a Failure. The code failed with a value of 'TError representing what went wrong.
      | Error of ErrorValue:'TError
```

The `Struct` attribute should be listed in the API reference signature.

The other two should be ignored.

PR #332, opened 2018-08-21
[WIP] F# signature gen for members, attributes, records, DUs, and Enums

cartermp:
Due to #308  I cannot test this (the test projects do not build for me), so here's to hoping that something useful comes out of CI!

Taking on:

#319 
#318 
#335
#337 
#339


cartermp:
@joelmartinez I've got this now to a point where I think it's good enough to do acceptance testing, but the build still fails after tests pass. Any ideas?

cartermp:
Ah durn, looks like there's some XML baselines that need updating. Can a tool do that?

joelmartinez:
@cartermp not really in this case (_some_ of those integration tests have ways to automatically update them in cases of breaking changes like this ... but it has to be done via CLI/make, and these f# tests in particular don't have those targets).

I'm working on a few high-pri things right now, but am going to try to make some time today or tomorrow to look at why this isn't building on windows, in addition to making the requisite edits to the test XMLs. Just grant me the ability to modify your branch:  
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

cartermp:
@joelmartinez according to the article you should have rights to edit the branch.

cartermp:
@joelmartinez I added a new file, and the project builds locally, but not on CI. The file is a part of these changes. Is there something else I need to do to make it build in CI?

cartermp:
never mind, it's old style csproj

joelmartinez:
@cartermp I'm now working on updating the baseline XML tests ... I've encountered one regression. Looks like you updated the generic parameter brackets to use `{` instead of `<`; however you've done so in a place that affected some of the other formatters (not sure offhand if it's the docid formatter, or the slashdoc formatter, in this case):

```
-      <Member Id="P:Mono.DocTest.Generic.MyList`2.System#Collections#Generic#ICollection&lt;A&gt;#IsReadOnly" />
-      <Member Id="P:Mono.DocTest.Generic.MyList`2.System#Collections#Generic#IEnumerator&lt;A&gt;#Current" />
+      <Member Id="P:Mono.DocTest.Generic.MyList`2.System#Collections#Generic#ICollection{A}#IsReadOnly" />
+      <Member Id="P:Mono.DocTest.Generic.MyList`2.System#Collections#Generic#IEnumerator{A}#Current" />
```

This would end up breaking a large amount of systems that depend on the current formatting for those IDs

mimisasouvanh:
[Internal work item](https://ceapex.visualstudio.com/Engineering/_workitems/edit/163531) created to track this bug

joelmartinez:
This work has unfortunately bit-rotted past the point where it is possible to reasonably re-integrate it with the latest code ... I'm going to close this PR, though we will keep it for reference if/when we have resources in the future to address some of these issues.

Issue #345, opened 2018-08-31
mdoc fails to document .NET assemblies


dend:
Assemblies: [VSTS](https://apidrop.visualstudio.com/_git/binaries?path=%2Fdotnet&version=GBnpe-2.1)

Latest build: [VSTS](https://apidrop.visualstudio.com/Content%20CI/_build/results?buildId=29399&view=logs)

mdoc release: 5.7.3

Error:

```bash
2018-08-31T22:50:56.6664185Z Updating: System.Net.CookieException
2018-08-31T22:50:56.6721348Z Updating: System.Net.CredentialCache
2018-08-31T22:50:56.6791656Z Updating: System.Net.DecompressionMethods
2018-08-31T22:50:56.6837627Z Updating: System.Net.DnsEndPoint
2018-08-31T22:50:56.6904498Z Updating: System.Net.EndPoint
2018-08-31T22:50:56.6985721Z Updating: System.Net.HttpStatusCode
2018-08-31T22:50:57.7898497Z ##[error]mdoc: System.InvalidCastException: Unable to cast object of type 'System.Xml.XmlComment' to type 'System.Xml.XmlElement'.
   at System.Linq.Enumerable.<CastIterator>d__97`1.MoveNext()
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at Mono.Documentation.MDocUpdater.WriteElement(XmlNode parent, String element, Boolean forceNewElement)
   at Mono.Documentation.MDocUpdater.UpdateType(XmlElement root, TypeDefinition type, FrameworkTypeEntry typeEntry)
   at Mono.Documentation.MDocUpdater.DoUpdateType2(String message, XmlDocument basefile, TypeDefinition type, FrameworkTypeEntry typeEntry, String output, Boolean insertSince)
   at Mono.Documentation.MDocUpdater.DoUpdateType(AssemblySet set, TypeDefinition type, FrameworkTypeEntry typeEntry, String basepath, String dest)
   at Mono.Documentation.MDocUpdater.DoUpdateAssembly(AssemblySet assemblySet, AssemblyDefinition assembly, XmlElement index_types, String source, String dest, HashSet`1 goodfiles)
   at Mono.Documentation.MDocUpdater.DoUpdateAssemblies(String source, String dest)
   at Mono.Documentation.MDocUpdater.Run(IEnumerable`1 args)
   at Mono.Documentation.MDoc.Main(String[] args)
[...]

PR #347, opened 2018-09-07
Now tracking all assemblies for forwarded/exported types.

### LINKED ISSUE #342: Retain full assembly information for forwarded types
As an example, `System.Net.WebResponse` is forwarded from `System.Net.dll`, therefore `System.Net.dll` is not included in the list of assemblies that show the origin of the API. The request is to include the full list of assemblies in the final ECMAXML output.

joelmartinez:
Resolves #342


joelmartinez:
@dend just fyi, I have this completed (for #347), but still need to add a unit test. Between that, and wanting to also include a fix for https://github.com/mono/api-doc-tools/issues/345, looks like this is going to slide in on Monday or Tuesday.

Issue #249, opened 2018-05-06
Incorrect MemberSignatures for ref returns and ref readonly returns


svick:
Consider the following C# 7.2 code:

```c#
public class C
{
    int i;

    public ref int Ref() => ref i;
    public ref readonly int ReadonlyRef() => ref i;
}
```

When I generate XML for it using `mdoc update -o en lib.dll`, the relevant parts of the output are:

```xml
<Member MemberName="ReadonlyRef">
  <MemberSignature Language="C#" Value="public Int32&amp; ReadonlyRef ();" />
  <MemberSignature Language="ILAsm" Value=".method public hidebysig instance class System.Int32&amp; modreq(System.Runtime.InteropServices.InAttribute) ReadonlyRef() cil managed" />
  <ReturnValue>
    <ReturnType>System.Int32&amp;</ReturnType>
  </ReturnValue>
</Member>
<Member MemberName="Ref">
  <MemberSignature Language="C#" Value="public int Ref ();" />
  <MemberSignature Language="ILAsm" Value=".method public hidebysig instance int32 Ref() cil managed" />
  <ReturnValue>
    <ReturnType>System.Int32@</ReturnType>
  </ReturnValue>
</Member>
```

The fact that `ReturnType` for `Ref` contains `@` has already been mentioned in https://github.com/mono/api-doc-tools/issues/230.

But what this issue is about are the `MemberSignature`s. I believe the member signatures for C# should use C# syntax for the two methods. That is, they should be:

```xml
<MemberSignature Language="C#" Value="public ref readonly int ReadonlyRef ();" />
<MemberSignature Language="C#" Value="public ref int Ref ();" />
```

dend:
@joelmartinez we'll likely need to bump this up, as it affects the correctness of signatures.

joelmartinez:
@dend ok, I'll cue this up for the next release

joelmartinez:
Implementation notes:

- These signatures are generated in the [`CSharpFullMemberFormatter`](https://github.com/mono/api-doc-tools/blob/master/mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs) class.
- You'll probably have to work in the `GetMethodDeclaration` class (which currently defers most of the work to the [base class' implementation](https://github.com/mono/api-doc-tools/blob/master/mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs#L374). 
[...]

PR #402, opened 2019-01-04
Fix MemberSignatures for ref returns and ref readonly returns

TianqiZhang:
This PR is to fix issue #249 . 

It seems that `type.IsByReference` is always false for return values, so I also updated the logic in MDocUpdater.cs.

Not sure whether we need integration test for this change, for now I only added one UT. @joelmartinez please let me know if I need to move the test case to somewhere else.

### DIFF
--- a/apidoctools.sln
+++ b/apidoctools.sln
@@ -15,8 +15,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "mdoc.Test", "mdoc\mdoc.Test
 EndProject
 Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "mdoc.Test.FSharp", "mdoc\mdoc.Test\mdoc.Test.FSharp\mdoc.Test.FSharp.fsproj", "{979F9F80-12FE-4236-9E93-6D554AB13701}"
 EndProject
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "mdoc.Test.Cplusplus", "mdoc.Test.Cplusplus\mdoc.Test.Cplusplus.vcxproj", "{9398D4E3-1779-44FD-AF8C-BB46562DCD88}"
-EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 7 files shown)


apidoctools.sln
```diff
@@ -13,12 +13,11 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ICSharpCode.SharpZipLib", "
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "mdoc.Test", "mdoc\mdoc.Test\mdoc.Test.csproj", "{5ADDEFB6-930C-46BC-8B2B-FDE5C7E3B5AD}"
 EndProject
-Project("{f2a71f9b-5d33-465a-a702-920d77279786}") = "mdoc.Test.FSharp", "mdoc\mdoc.Test\mdoc.Test.FSharp\mdoc.Test.FSharp.fsproj", "{979F9F80-12FE-4236-9E93-6D554AB13701}"
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "mdoc.Test.FSharp", "mdoc\mdoc.Test\mdoc.Test.FSharp\mdoc.Test.FSharp.fsproj", "{979F9F80-12FE-4236-9E93-6D554AB13701}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "mdoc.Test.Cplusplus", "mdoc.Test.Cplusplus\mdoc.Test.Cplusplus.vcxproj", "{9398D4E3-1779-44FD-AF8C-BB46562DCD88}"
```
joelmartinez: Unfortunately, adding this project to the `apidoctools.sln` makes it so that you can't build in VSMac … as frustrating as it surely is when edits need to be made to the cpp project, it should be removed for the purposes of max multi-platform friendliness for maintainers.

apidoctools.sln
```diff
@@ -126,13 +125,28 @@ Global
...
+		{9398D4E3-1779-44FD-AF8C-BB46562DCD88}.Release|x86.Build.0 = Release|Win32
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {6D6A6927-89D3-42DF-A416-294A841822A9}
 	EndGlobalSection
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
```
joelmartinez: avoid adding this … as it's not really something that works on VSMac

mdoc/Makefile
```diff
@@ -48,7 +48,7 @@ MULTI-UNIFIED = Test/DocTest-DropNS-unified.dll Test/DocTest-DropNS-unified-mult
 
 DIFF = diff -rup
 DIFF_QUIET = diff --brief
-ifeq ($(PLATFORM), win32)
+ifeq ($(OS), Windows_NT)
```
joelmartinez: were you able to get the makefile building on windows? if so that's *awesome*! Can you update [the readme](https://github.com/mono/api-doc-tools#cli) to include any specifics on how you got this working (including what version of `make` you installed)?

mdoc/mdoc.Test/FormatterTests.cs
```diff
@@ -207,6 +207,15 @@ public void IL_RefAndOut ()
                 "~SomeGenericClass ();",
                 "Finalize");
 
+        [Test]
+        public void CSharpReadonlyRef()
+        {
+            var member = GetMethod(typeof(ReadonlyRefClass), m => m.Name == "ReadonlyRef");
+            var formatter = new CSharpFullMemberFormatter();
+            var sig = formatter.GetDeclaration(member);
+            Assert.AreEqual("public ref readonly int ReadonlyRef ();", sig);
```
joelmartinez: This unit test is great … but I'd like to see a few more combinations of unit tests, such as a method that returns only `ref` (I _believe_ you can't have only a `readonly` returns, right?). But in any case, unit tests for every variation would be ideal.

apidoctools.sln
TianqiZhang: I see...it's removed now.

apidoctools.sln
TianqiZhang: removed

mdoc/mdoc.Test/FormatterTests.cs
TianqiZhang: good catch! Added another case for `ref` return only, and found a bug in my change 😉 . Updated.

mdoc/Makefile
joelmartinez: ok that makes sense

PR #433, opened 2019-05-03
Suppress "put_" property setters in mdoc for UWP components.

IrinVoso:
UWP components use the prefix "put_" for property setter methods. mdoc does not currently recognize this prefix and thus creates an additional individual method for the property setter.

Internal bug: [76428](https://ceapex.visualstudio.com/Engineering/_workitems/edit/76428)


dendeli-msft:
@joelmartinez does this also need a change in Mono.Cecil?

joelmartinez:
@dendeli-msft no, this is mdoc-specific logic … basically we just need to ignore the "special" internally generated functions, and in this case, UWP simply deviated from the special names used by roslyn C# compiler :)

external/Test/readme.md
```diff
@@ -0,0 +1,10 @@
+The files in this directory are used as test inputs for explicit assembly testing.
```
joelmartinez: Great idea adding this readme

joelmartinez:
@IrinVoso … I'll let you know when this makes it into a release. I hope to be able to compile this along with a few other changes into a release soon 🎉

PR #438, opened 2019-08-05
Required updates to work with new Cecil (not yet released)

marek-safar:


marek-safar:
Cecil releases are made by @jbevain who can comment on that

PR #440, opened 2019-08-12
Fix signatures for ref return indexers

svick:
Fixes https://github.com/mono/api-doc-tools/issues/422.

Also cleans up nearby code.

Is there some easy way to see the diff caused by this PR on the libraries this project is used for? (To make sure it doesn't cause any accidental changes.)


svick:
@joelmartinez Sorry it took me so long to get back to this. I have now merged in master and addressed issues I know about (including the `ref ref` issue you mentioned).

[After I locally fixed some tests that already fail for me on master](https://github.com/mono/api-doc-tools/compare/master...0e6b2c1), all tests (`make prepare all check`) succeed for me. And I can't access CI results, to see why it's failing now.

joelmartinez:
@svick hmm ... interesting, it's one of the unit tests that's failing. You should be able to run these locally, even in visual studio ([instructions here](https://github.com/mono/api-doc-tools/blob/master/CONTRIBUTING.md#unit-tests))

This is the details on the test that fails, so it looks like the test expectation just needs to be adjusted for your new changes:
```
1) Failed : mdoc.Test.FormatterTests.FuncParams
  Expected string length 74 but was 75. Strings differ at index 39.
  Expected: "...meMethod4 (out string a, in int i, T t, object b = default);"
  But was:  "...meMethod4 (out string a, ref int i, T t, object b = default);"
  ---------------------------------------^
  at mdoc.Test.FormatterTests.FuncParams () [0x00000] in <33362d4116d6487a824a74180729d3c3>:0 
```

svick:
@joelmartinez I don't understand that. That test should be correct, since the corresponding parameter (which was added by me) is `in`:

https://github.com/mono/api-doc-tools/blob/4b2f1b215703bf9805de3b6e0b0de052bf67f678/mdoc/mdoc.Test/SampleClasses/SomeGenericClass.cs#L17

And the test does pass for me both in VS and when using `make check` on WSL. I was wondering if the difference could be caused by different IL, so I expanded the test to also check that (and also cleaned up some related code). But if that doesn't show anything, I don't know what the difference between my machine and CI could be. Can you think of anything?

(While looking into this, I realized `in` parameters are not the same thing as `[In] ref` parameters, hence the commit ca92745, but I'm assuming that didn't affect this failure.)

joelmartinez:
@svick hmm, I'm not sure what might be happening here ... the current failure is still on that same test but now it's just missing the modifier:  
image

Really odd that it passes locally for you, but not on the CI 🤔 Surely that means there's a mismatch between what the CI is running and what you're running locally? The CI is running on the hosted macos agents

image

Some quick searching shows that this is what's installed on those agents:  
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

Can you compare and see what version of mono you have installed locally compared to that?

svick:
@joelmartinez 

> the current failure is still on that same test but now it's just missing the modifier

Yeah, that's the same failure, I have just expanded the test a bit in an attempt to narrow down the issue.

> Can you compare and see what version of mono you have installed locally compared to that?

I'm using the latest version of Mono that's available on Ubuntu 20.04: 6.10.0.104.

For the record, the Dockerfile I'm using is:

```dockerfile
FROM ubuntu:20.04

RUN apt-get update
RUN apt-get install gnupg2 -y

RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
RUN echo "deb http://download.mono-project.com/repo/ubuntu focal main" | tee /etc/apt/sources.list.d/mono-official.list

RUN apt-get update
ENV TZ=Europe/Prague
RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
RUN apt-get install tzdata nuget -y --allow-unauthenticated
RUN apt-get install git make mono-devel mono-vbnc ca-certificates-mono wget fsharp -y

RUN wget -q 'https://hg.mozilla.org/mozilla-central/raw-file/tip/security/nss/lib/ckfw/builtins/certdata.txt' -O "/tmp/certdata.txt"
RUN mozroots --import --ask-remove --file /tmp/certdata.txt

ARG CACHE_BUST

RUN git clone https://github.com/svick/api-doc-tools
WORKDIR api-doc-tools
RUN make prepare all check
```

And I run it using `docker build . -t mdoc --build-arg "CACHE_BUST=$(date +%s)"`.

huangmin-ms:
Close this PR as the issue is addressed in https://github.com/mono/api-doc-tools/pull/611

PR #443, opened 2019-10-05
mdoc 5.7.4.12

joelmartinez:
Proposal for metadata schema


mdoc/Resources/monodoc-ecma.xsd
```diff
@@ -485,6 +501,7 @@ add masterdoc support?
   <xs:element name="Member">
     <xs:complexType>
       <xs:sequence>
+        <xs:element ref="Meta" minOccurs="0" maxOccurs="1" />
```
daxianji007: should this be "Metadata" instead of "Meta"?

joelmartinez: Yes indeed! great catch ... thanks for the review :)

PR #445, opened 2019-12-12
REPLACE "NULL" TO "DEFAULT"

maopeixia:


joelmartinez:
closing and re-opening the pull request to get the CI to pick up and queue a build :)

joelmartinez:
once again ... had to add `marj-metadata` to the list of branches the CI recognizes

mdoc/Mono.Documentation/Updater/Formatters/AttributeValueFormatter.cs
```diff
@@ -1,66 +1,66 @@
...
-    }
+    /// <summary>Formats attribute values. Should return true if it is able to format the value.</summary>
+    class AttributeValueFormatter
+    {
+        public virtual bool TryFormatValue (object v, ResolvedTypeInfo type, out string returnvalue)
+        {
+            TypeReference valueType = type.Reference;
+            if (v == null)
+            {
+                returnvalue = "default";
```
joelmartinez: ok ... so on first glance, this looks fantastic, great job 👏 Unfortunately on review I'm realizing that this fixes the issue in one language while introducing a regression for the other languages 😊

`default` is only a C# language construct ... it doesn't exist in the other languages. So instead of making the change here in this class, you should make the change in the `CSharpFullMemberFormatter` class. Look in the `AppendParameter` method there, and you will find a line that calls `MDocUpdater.MakeAttributesValueString`. You can probably add the logic in that section to call that method, and if it's null, switch it to `default`.

joelmartinez:
Oh, and also ... adding that method to the test class, caused some follow up changes to the integration changes. I recommend getting those running, so that you can fix this and any other items that come up from integration tests using this same test class:

```
+    <Member MemberName="DoWithNullParams">
+      <MemberSignature Language="C#" Value="public string DoWithNullParams (out string a, object b = default, mdoc.Test.SampleClasses.TestClass c = default);" />
+      <MemberSignature Language="ILAsm" Value=".method public hidebysig instance string DoWithNullParams([out] string&amp; a, object b, class mdoc.Test.SampleClasses.TestClass c) cil managed" />
+      <MemberType>Method</MemberType>
+      <AssemblyInfo>
+        <AssemblyVersion>0.0.0.0</AssemblyVersion>
+      </AssemblyInfo>
+      <ReturnValue>
+        <ReturnType>System.String</ReturnType>
+      </ReturnValue>
+      <Parameters>
+        <Parameter Name="a" Type="System.String" RefType="out" />
+        <Parameter Name="b" Type="System.Object" />
+        <Parameter Name="c" Type="mdoc.Test.SampleClasses.TestClass" />
+      </Parameters>
+      <Docs>
+        <param name="a">To be added.</param>
+        <param name="b">To be added.</param>
+        <param name="c">To be added.</param>
+        <returns>To be added.</returns>
+      </Docs>
+    </Member>
```

PR #450, opened 2019-12-27
Solve the Math.PI syntax displays wrong value

maopeixia:


mdoc/Mono.Documentation/Updater/DocUtils.cs
```diff
@@ -412,7 +412,11 @@ public static void AppendFieldValue(StringBuilder buf, FieldDefinition field)
                     buf.Append(" = ").Append(val.ToString());
                 else if (val is IFormattable)
                 {
-                    string value = ((IFormattable)val).ToString(null, CultureInfo.InvariantCulture);
+                    string value = "";
+                    if (field.FieldType.FullName == "System.Double")
+                    { value = ((IFormattable)val).ToString("R", CultureInfo.InvariantCulture); }
```
tannergooding: I don't think using `R` is necessarily appropriate.

On Full Framework, there is no guarantee that `R` provides a string that will actually roundtrip (due to issues in the floating-point parser that can't be fixed due to compat reasons, etc; for .NET Core 3.0 and later, `R` and `G` will always roundtrip).

The appropriate thing is to likely use `G17` which will produce enough digits to roundtrip. `17` is the maximum precision supported by full framework (.NET core supports up to `99`). `G17` will always produce a roundtrippable string on .NET Core and will in almost every scenario produce a roundtrippable string on full framework.

`G9` is the precision that should be used for `System.Single`

tannergooding: I don't know about the guarantees of Mono; but I believe it is sharing the parsing/formatting code with .NET Core; so it likely depends on the version of Mono being used.

tannergooding:
Separately from changing the value string displayed in the documentation; we probably want a short snippet for core values (like `Math.PI`) explaining why the value shown to 17 digits isn't the "expected" value.

For example, the actual values are approx:
```
MathF.PI:  3.14159274...
Math.PI:   3.1415926535897931...
Actual PI: 3.1415926535897932384626433...
```

Some people may expect `MathF.PI` to end in `65` rather than `74` and for `Math.PI` to end in `32` rather than `31`; but float/double are merely the nearest representable value.

joelmartinez:
@tannergooding so @maopeixia has made this change, however one of the existing tests is having a regression. As you can [see here](https://github.com/mono/api-doc-tools/blob/master/mdoc/Test/DocTest-v1.cs#L209), there was an existing test that used an approximation of PI (how appropriate :P ) as a test value:

    protected internal const double PI = 3.14159;

Unfortunately, as you can see here in the diff ... using `G17` for `double` results in the following changed values:

```
-      <MemberSignature Language="C#" Value="protected internal const double PI = 3.14159;" />
-      <MemberSignature Language="ILAsm" Value=".field familyorassembly static literal float64 PI = (3.14159)" />
+      <MemberSignature Language="C#" Value="protected internal const double PI = 3.1415899999999999;" />
+      <MemberSignature Language="ILAsm" Value=".field familyorassembly static literal float64 PI = (3.1415899999999999)" />
```
Not exactly a more accurate result :| I'd like to propose that we go back to using `R` for both double and single, and let newer releases of mono/framework catch up to that new .net core 3 implementation over time. Are you ok with that?

maopeixia:
I've went back to using "R" format.

tannergooding:
> Not exactly a more accurate result

Actually, it's significantly more accurate. It shows that `3.14159` is not actually `3.14159` and is instead `3.14158999999999988261834005243144929409027099609375` (due to the original value not being exactly representable).

You ultimately need to decide between having accurate roundtrippable numbers that may be overly verbose or returning values that are "pretty" but which may have off by one (or more) errors when parsed.

On .NET Framework and .NET Core prior to 3.0, using `G` is equivalent to `G15`. Using `R` is equivalent to attempting `G15`, parsing to see if the value roundtrips and using `G17` if that fails. However, due to bugs in the parser, this isn't guaranteed to produce a roundtrippable result and may incorrectly truncate certain values.

On .NET Core 3.0 and later, `G` is equivalent to `R` and both always produce the shortest roundtrippable string.

A (non-exhaustive) selection of the issues that were resolved due to the above issues and which go into more detail are here:
https://github.com/dotnet/coreclr/issues/1316, https://github.com/dotnet/coreclr/issues/3313, https://github.com/dotnet/coreclr/issues/13106, https://github.com/dotnet/coreclr/issues/13615, https://github.com/dotnet/coreclr/issues/17467, https://github.com/dotnet/coreclr/issues/19802, https://github.com/dotnet/corefx/issues/31847, etc

joelmartinez:
Closing and re-opening to kick off CI

Issue #399, opened 2018-12-24
<AssemblyVersion> missing in <AssemblyInfo> tag


TianqiZhang:
Take `M:System.Data.SqlClient.SqlDataReader.GetColumnSchema` for example, you can see in its [ECMAXML file](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Data.SqlClient/SqlDataReader.xml#L606), there's a assembly info tag that does not have any versions at all:

```xml
      <AssemblyInfo>
        <AssemblyName>System.Data</AssemblyName>
      </AssemblyInfo>
```

joelmartinez:
This will *probably* be mitigated in #373, but leaving open to verify

joelmartinez:
I've opened a separate [internal issue](https://ceapex.visualstudio.com/Engineering/_workitems/edit/163956) for this particular item. /cc @mimisasouvanh

PR #452, opened 2020-01-06
Add logic to generate type forwarding information to ECMAXML

anmeng10101:


anmeng10101:
@joelmartinez 

I reused the methods in [FXUtils](https://github.com/mono/api-doc-tools/blob/master/mdoc/Mono.Documentation/Updater/Frameworks/FXUtils.cs) to add/remove fx to/from list.

Since the changes will be cached, I am not sure it will impact the results return by [PreviouslyProcessedFXString](https://github.com/mono/api-doc-tools/blob/marj-metadata/mdoc/Mono.Documentation/MDocUpdater.cs#L3896).

joelmartinez:
Closing and re-opening to request new build after small update in develop branch for failing unit test

mdoc/Test/en.expected.typeForwards/TheNamespace/TheClass.xml
```diff
@@ -12,6 +12,17 @@
...
   </AssemblyInfo>
+  <AssemblyInfo>
+    <AssemblyName>DocTest-typeForwards-Third</AssemblyName>
+    <AssemblyVersion>0.0.0.0</AssemblyVersion>
+  </AssemblyInfo>
+  <AssemblyInfo>
+    <AssemblyName>DocTest-typeForwards-Third-First</AssemblyName>
+  </AssemblyInfo>
+  <TypeForwardingChain>
+    <TypeForwarding From="DocTest-typeForwards-Second-First" FromVersion="0.0.0.0" To="DocTest-typeForwards-Second" ToVersion="0.0.0.0" />
```
joelmartinez: Isn't this the type forward for the framework `Two` ? This FrameworkAlternate attribution doesn't seem correct. If I'm remembering this test correctly, I would have expected to see `FrameworkAlternate="Two"` on this one, and `FrameworkAlternate="Three"` on the following line ... right?

mdoc/Test/en.expected.typeForwards/TheNamespace/TheClass.xml
```diff
@@ -12,6 +12,17 @@
   <AssemblyInfo>
     <AssemblyName>DocTest-typeForwards-Second-First</AssemblyName>
   </AssemblyInfo>
+  <AssemblyInfo>
+    <AssemblyName>DocTest-typeForwards-Third</AssemblyName>
+    <AssemblyVersion>0.0.0.0</AssemblyVersion>
+  </AssemblyInfo>
+  <AssemblyInfo>
+    <AssemblyName>DocTest-typeForwards-Third-First</AssemblyName>
```
joelmartinez: This isn't strictly in the scope of this particular task, but since you're already going to look at it again for the other comment ... can you find out why this doesn't have an `AssemblyVersion` ? Making sure that we properly log all the AssemblyVersion information would go a long way to helping us figure out resolution issues when they occur :)

mdoc/Test/DocTest-typeForwards.cs
```diff
@@ -3,6 +3,11 @@
 [assembly:TypeForwardedToAttribute(typeof(TheNamespace.TheClass))] 
 #endif
 
+#if THIRD
```
joelmartinez: Not a huge deal, but strictly speaking this is the same as the `TWO` ... so you could simplify this declaration by doing

```
#if SECOND || THIRD
```

mdoc/Test/en.expected.typeForwards/TheNamespace/TheClass.xml
anmeng10101: To make sure the test case will cover removing "FrameworkAlternate" attribute if it includes all frameworks, I edited "make" file to add type forwarding for framework `One` as well.

```
cp Test/DocTest-typeForwards-Second-First.dll Test/FrameworkTestData-fx-typeForwards/One
cp Test/DocTest-typeForwards-Second.dll Test/FrameworkTestData-fx-typeForwards/dependencies/One
```

mdoc/Test/DocTest-typeForwards.cs
anmeng10101: Thanks for the suggestion, improved

mdoc/Test/en.expected.typeForwards/TheNamespace/TheClass.xml
anmeng10101: Good catch, also noticed `AssemblyVersion` missing both for type and members of the "from" assembly(such as DocTest-typeForwards-Second-First).

Did some investigation, this could related to the lines here

[For Type](https://github.com/mono/api-doc-tools/blob/develop/mdoc/Mono.Documentation/MDocUpdater.cs#L2219)
`
Func<XmlElement, bool> assemblyFilter = x => x.SelectSingleNode ("AssemblyName").InnerText == type.Module.Assembly.Name.Name;
`

Take DocTest-typeForwards-Second-First.dll for example, the type.module and member.module will return "DocTest-typeForwards-Second.dll".

So the node `<AssemblyName>DocTest-typeForwards-Second-First</AssemblyName>` will never be picked up to update assembly version.

I think we may need another bug to track the issue and discuss the fix of it.

type-module

joelmartinez: There *is* a GH bug which discusses this very issue (#399) ... but no internal item, so I've [created one](https://ceapex.visualstudio.com/Engineering/_workitems/edit/163956) that we can address separately :) /cc @mimisasouvanh

mdoc/Test/en.expected.typeForwards/TheNamespace/TheClass.xml
joelmartinez: I think this is still not right ... the `One` framework shouldn't have this assembly (as it kind of negates the entire test, which is meant to represent a type being introduced in one version, then being moved into another assembly in a future release).

If you remove this most-recent change (so that the type forward exists only in `Two` and `Three`) ... then I would expect to see the following in the XML:

```xml
<TypeForwarding From="DocTest-typeForwards-Second-First" FromVersion="0.0.0.0" To="DocTest-typeForwards-Second" ToVersion="0.0.0.0" FrameworkAlternate="Two />
<TypeForwarding From="DocTest-typeForwards-Third-First" FromVersion="0.0.0.0" To="DocTest-typeForwards-Third" ToVersion="0.0.0.0" FrameworkAlternate="Three" />
```
Which says, "in moniker `Two`, it's forwarded from X to Y, and in moniker `Three` it's forwarded from moniker Z to N" ... of course, if a type is forwarded in *every moniker*, then I would expect to see the `TypeForward` tag but with no `FrameworkAlternate` attribute, because at that point the list of frameworks will be all of them :)

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -3100,11 +3102,109 @@ private static bool GetFieldConstValue (FieldDefinition field, out string value)
...
+            }
+
+            return exsitingChain;
         }
 
-        // XML HELPER FUNCTIONS
+        /// <summary>
+        /// Purge FrameworkAlternate attribute if it includes all frameworks 
+        /// </summary>
+        private void PurgeFrameworkAlternateAttribute(FrameworkTypeEntry typeEntry, IEnumerable<XmlElement> nodes)
```
joelmartinez: Maybe this method could be moved into DocUtils ?

PR #453, opened 2020-01-07
Append framework to "FrameworkAlternate" attribute if base type name exists

anmeng10101:


anmeng10101:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/156999/

joelmartinez:
@anmeng10101 before I merge this into the develop branch ... are there any related changes that need to happen downstream, like in Ecma2Yaml to support this change?

joelmartinez:
closing and re-opening to request new build after small test update in develop branch

mdoc/Test/DocTest-framework-inheritance.cs
```diff
@@ -18,6 +18,9 @@ public class MyClass
     #if FXTWO
         : MyBaseClassTwo
     #endif
+    #if FXTHREE
```
joelmartinez: this could have been simplified by simply adding `|| FXTHREE` to the first conditional preprocessor directive ... but it's not a huge deal :)

PR #455, opened 2020-01-13
EII url missing namespace in the middle

maopeixia:
@joelmartinez 
orginal xml:
image

now xml
image

I solved the array problem first,then check the parameters of other collection types.


maopeixia:
Code:
return type.FullName.Replace(elements.Name, genericTypes[elements.Name])
Ci return value is inconsistent with local test return value. CI error, I'll solve the problem tomorrow

maopeixia:
@joelmartinez 
Sorry, a lot of commits has been submitted,Because CI requires a lot of file change ,otherwise errors.

PR #458, opened 2020-01-17
generate type signature per base types with "FrameworkAlternate"

anmeng10101:


joelmartinez:
Looks like this CI run failed due to the result of your test change not being exactly as you expected:

```
--- Test/en.expected-frameworks-inheritance/MyNamespace/MyClass.xml	2020-01-17 05:56:58.000000000 -0500
+++ Test/en.actual/MyNamespace/MyClass.xml	2020-01-17 05:58:18.000000000 -0500
@@ -1,8 +1,8 @@
 <Type Name="MyClass" FullName="MyNamespace.MyClass">
   <TypeSignature Language="C#" Value="public class MyClass : MyNamespace.MyBaseClassOne" FrameworkAlternate="One" />
   <TypeSignature Language="ILAsm" Value=".class public auto ansi beforefieldinit MyClass extends MyNamespace.MyBaseClassOne" FrameworkAlternate="One" />
-  <TypeSignature Language="C#" Value="public class MyClass : MyNamespace.MyBaseClassTwo" FrameworkAlternate="Two" />
-  <TypeSignature Language="ILAsm" Value=".class public auto ansi beforefieldinit MyClass extends MyNamespace.MyBaseClassTwo" FrameworkAlternate="Two" />
+  <TypeSignature Language="C#" Value="public class MyClass : MyNamespace.MyBaseClassTwo" FrameworkAlternate="Three;Two" />
+  <TypeSignature Language="ILAsm" Value=".class public auto ansi beforefieldinit MyClass extends MyNamespace.MyBaseClassTwo" FrameworkAlternate="Three;Two" />
```

anmeng10101:
When I run smoke test locally, pr #452 not merge to develop yet 😁

Let me pull latest develop branch and update expected test result.

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -2518,25 +2518,80 @@ private void UpdateMember (DocsNodeInfo info, FrameworkTypeEntry typeEntry, Dict
...
+                if (!string.IsNullOrWhiteSpace(valueToUse))
+                    newElement.SetAttribute("Value", valueToUse);
+
+                if (!string.IsNullOrWhiteSpace(usageSample))
+                    newElement.SetAttribute("Usage", usageSample);
+
+                newElement.SetAttribute(Consts.FrameworkAlternate, typeEntry.Framework.Name);
+            }
+
+            // if last framework for type, check if need to purge FrameworkAlternate attribute
```
anmeng10101: @joelmartinez, I will create a separate pr to move method `PurgeFrameworkAlternateAttribute` to DocUtils.cs, change this section of code to use generic method.

joelmartinez: ok

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -2518,25 +2518,80 @@ private void UpdateMember (DocsNodeInfo info, FrameworkTypeEntry typeEntry, Dict
...
+            var valueToUse = formatter.GetDeclaration(type);
+            if (valueToUse == null && usageSample == null)
+                return;
+
+            bool elementFound = false;
+
+            // if there is already signature found under same formatter, update fxa
+            if (existingElements.Any())
+            {
+                foreach (var element in existingElements)
```
joelmartinez: ok, so I'm comparing this to the similar `MemberSignature` method, and it looks like this particular section isn't complete. This is what _that_ section looks like:

```csharp
                foreach(var element in existingElements)
                {
                    var val = element.GetAttribute("Value");
                    var usage = element.GetAttribute("Usage");
                    if (val == (valueToUse??"") && usage == (usageSample??""))
                    {
                        // add FXA
                        string newfxa = FXUtils.AddFXToList(element.GetAttribute(Consts.FrameworkAlternate), typeEntry.Framework.Name);
                        element.SetAttribute(Consts.FrameworkAlternate, newfxa);
                        elementFound = true;
                    }
                    else
                    {
                        string newfxa = FXUtils.RemoveFXFromList(element.GetAttribute(Consts.FrameworkAlternate), typeEntry.Framework.Name);
                        element.SetAttribute(Consts.FrameworkAlternate, newfxa);
                    }
                }
```

Note the _else_ branch, which will remove the framework from that element (if it's there) if `val` and `usage` don't match.

Actually ... now that I'm working through this, I wonder if this is just an artifact of the previous iteration where it wasn't clearing the list every time. Is this `else` in the other method necessary?

joelmartinez: as discussed ... this `else` is not necessary

PR #471, opened 2020-02-28
Incorrect signatures for span indexers

maopeixia:
The unit test  uses the class written by myself, At the same time, it has also been tested with struct span < T > in  system.memory.dll,It also passed.

bug:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/167409
https://ceapex.visualstudio.com/Engineering/_workitems/edit/167407

### DIFF
--- a/mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
+++ b/mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
@@ -598,6 +598,9 @@ namespace Mono.Documentation.Updater
                 modifiers = "";
             buf.Append (modifiers).Append (' ');
 
+            if (property.PropertyType.IsByReference)
+                buf.Append("ref ");
+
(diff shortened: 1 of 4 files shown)


maopeixia:
As you requested ,Add a test using a ref struct,But I want to make it clear that this is different from the reality(struct span<T> in system.memory) .
Results of Decompilation:

it uses a pointer,but  T* Not supported in C# first,I try to use(int*) restore the code of this structure. Unfortunately, there's something wrong with not succeeding.Maybe this is a DLL written in C,which is different from C#.

maopeixia:
As you requested ,Add a test using a ref struct,But I want to make it clear that this is different from the reality(struct span<T> in system.memory) .
Results of Decompilation:
image
it uses a pointer,but  T* Not supported in C# first,I try to use(int*) restore the code of this structure. Unfortunately, there's something wrong with not succeeding.Maybe this is a DLL written in C,which is different from C#.

PR #477, opened 2020-03-13
Some classes are missing implicit conversion operators

maopeixia:
@joelmartinez 
Bug: https://ceapex.visualstudio.com/Engineering/_workitems/edit/167411
  This bug, I don't know how to write unit tests，But after the code is modified, I have tested it completely locally By frameworks.xml，Please let me know if you have any suggestions。

### DIFF
--- a/mdoc/Mono.Documentation/MDocUpdater.cs
+++ b/mdoc/Mono.Documentation/MDocUpdater.cs
@@ -1734,6 +1734,9 @@ namespace Mono.Documentation
                             }
                         }
 
+                        if (DocUtils.DocIdCheck(seenmembers[sig], oldmember))
+                        { continue; }
+
                         oldmember.SetAttribute ("ToDelete", "true");
                         DeleteMember ("Duplicate Member Found", output, oldmember, todelete, type);
                         statisticsCollector.AddMetric(typeEntry.Framework.Name, StatisticsItem.Members, StatisticsMetrics.Removed);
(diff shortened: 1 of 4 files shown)


mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -1689,9 +1689,27 @@ public void DoUpdateType2 (string message, XmlDocument basefile, TypeDefinition
                     DeleteMember ("Member Removed", output, oldmember, todelete, type);
                     statisticsCollector.AddMetric(typeEntry.Framework.Name, StatisticsItem.Members, StatisticsMetrics.Removed);
                     continue;
-                }
+                }
+
+                // Duplicated
+                   //CheckFunction
+                    Func<XmlNode, XmlElement, Boolean> Dochk = (a, b) =>
```
joelmartinez: Please change the name of this variable to something a bit more descriptive ... something like `docIdCheck` works :)

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -1689,9 +1689,27 @@ public void DoUpdateType2 (string message, XmlDocument basefile, TypeDefinition
...
+                    {
+                        if (b.LocalName != "Member" || a.LocalName != "Member")
+                            return false;
+
+                        var oldMembersDocid = b.SelectSingleNode("MemberSignature[@Language='DocId']");
+                        var seenNembersDocid = a.SelectSingleNode("MemberSignature[@Language='DocId']");
+
+                        if (oldMembersDocid != null && seenNembersDocid != null)
+                        {
+                            if (!seenNembersDocid.OuterXml.Equals(oldMembersDocid.OuterXml))
```
joelmartinez: Instead of checking `.OuterXml`, which as I understand it isn't that fast, please just compare the `Value` attribute (with proper null checking, of course) ... that should be a bit more performant :)

maopeixia: The code has been modified as you requested And Unit tests have been added。

PR #479, opened 2020-03-18
add type forwarding for multiple levels nested type

anmeng10101:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/185167

1. For inner type condition, `IsNested = true` is enough.
2. There could be more then 2 levels of nested, for example Dictionary`2+KeyCollection+Enumerator. So we need to handle check and apply forwarding information for each layer.

### DIFF
--- a/mdoc/Mono.Documentation/MDocUpdater.cs
+++ b/mdoc/Mono.Documentation/MDocUpdater.cs
@@ -3217,10 +3217,19 @@ namespace Mono.Documentation
 
             forwardings.AddRange(getForwardings(type));
 
-            // If current type is nested inner class, add forwardings from outer class as well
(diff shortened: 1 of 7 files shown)


anmeng10101:
@joelmartinez commit collapsed.

The recursive nested types have been considered and taken care in this [while](https://github.com/mono/api-doc-tools/pull/479/files#diff-6ebea535d3ba2330a8fcf26b35409a8fR3228) loop.

Here is an example for 3 levels nested type, forwarding was added after apply this fix

https://github.com/dotnet/dotnet-api-docs/blob/anmeng-185167/xml/System.Collections.Generic/Dictionary%602+KeyCollection+Enumerator.xml#L29

PR #490, opened 2020-04-03
Adding FrameworkAlternate to ReturnType

joelmartinez:
for:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/191858


TianqiZhang:
@joelmartinez should we also update `monodoc-ecma.xsd` to reflect this change?

PR #494, opened 2020-04-23
Attached properties are missing some versions on Applies to section

maopeixia:
Use the FRAMEWORK 4,7 version of presentationcore.dll for testing is successful.
image

### DIFF
--- a/mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs
+++ b/mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs
@@ -122,12 +122,12 @@ namespace Mono.Documentation.Util
             // https://github.com/mono/api-doc-tools/issues/63#issuecomment-328995418
             if (!field.Name.EndsWith(PropertyConst, StringComparison.Ordinal))
                 return false;
-            var propertyName = GetPropertyName(field.Name);
+            var propertyName = GetPropertyName(field.Name); 
             var getMethodName = $"Get{propertyName}";
             var setMethodName = $"Set{propertyName}";
 
-            var hasExistingProperty = field?.DeclaringType?.Properties.Any (p => p.Name.Equals (propertyName, System.StringComparison.Ordinal));
-            var hasExistingField = field?.DeclaringType?.Fields.Any (f => f.Name.Equals (propertyName, System.StringComparison.Ordinal));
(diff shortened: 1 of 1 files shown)


mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs
```diff
@@ -126,7 +126,7 @@ private static bool IsAttachedProperty(FieldDefinition field, Dictionary<string,
             var getMethodName = $"Get{propertyName}";
             var setMethodName = $"Set{propertyName}";
 
-            var hasExistingProperty = field?.DeclaringType?.Properties.Any (p => p.Name.Equals (propertyName, System.StringComparison.Ordinal));
+            var hasExistingProperty = field?.DeclaringType?.Properties.Any (p => p.Name.Equals (propertyName, System.StringComparison.Ordinal) && ChkFieldVisible(p));
             var hasExistingField = field?.DeclaringType?.Fields.Any (f => f.Name.Equals (propertyName, System.StringComparison.Ordinal));
```
joelmartinez, line 130: This line should have a similar check to the property check ... so once fully changed, it should look like this:

```csharp
            var hasExistingProperty = field?.DeclaringType?.Properties.Any (p => p.Name.Equals (propertyName, System.StringComparison.Ordinal) && ChkPropertyVisible(p));
            var hasExistingField = field?.DeclaringType?.Fields.Any (f => f.Name.Equals (propertyName, System.StringComparison.Ordinal) && ChkFieldVisible(f));
```

mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs
```diff
@@ -180,6 +180,28 @@ private static bool IsAssignableTo(TypeReference type, string targetTypeName)
             return type.FullName == targetTypeName || IsAssignableTo(typeDefenition.BaseType, targetTypeName);
         }
 
+        private static bool ChkFieldVisible(PropertyDefinition field)
```
joelmartinez, line 183: This should still be called `ChkPropertyVisible` ... and you should add **another**, separate, method called `ChkFieldVisible` to service line 130 above. This new method should accept a `FieldDefinition`, and it won't have the same implementation (it will be much simpler) as the property one ... because a field is just a field, so you don't have to check any other methods. You can basically just verify to see if the field itself is public or is explicitly implemented

PR #495, opened 2020-04-29
[mdoc] AssemblyVersion missing in some instances

maopeixia:
bug: https://ceapex.visualstudio.com/Engineering/_workitems/edit/163956/


mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -1633,7 +1633,15 @@ public void DoUpdateType2 (string message, XmlDocument basefile, TypeDefinition
                         oldmember.SetAttribute ("MemberName", memberName);
                     }
 
-                    AddEiiNameAsAttribute(info.Member, oldmember, memberName);
+                    AddEiiNameAsAttribute(info.Member, oldmember, memberName);
+
+                    // Check valid AssemblyInfo and delete
+                    if (typeEntry.Framework.IsFirstFrameworkForType(typeEntry))
```
joelmartinez, line 1639: So this check here is good ... this will take care of this logic at the member level. However, we still have to check it at the type level :) 

You can probably add that in the UpdateType function, before mdoc adds assemblies, here
https://github.com/mono/api-doc-tools/blob/483c3fca179bbc4ade06417bf2055c0f953c89d3/mdoc/Mono.Documentation/MDocUpdater.cs#L2243

maopeixia:
@joelmartinez 
I've changed the Code。

mdoc/Mono.Documentation/MDocUpdater.cs
joelmartinez: This right here ... this should be added back :)

maopeixia:
@joelmartinez 
Sorry I didn't explain yesterday's revision to you. I removed it from the member level , Because Pass the test, code : ".GetElementsByTagName("AssemblyInfo").Cast<XmlElement>()"
Type level already contains < AssemblyInfo > under type node and < AssemblyInfo > under member node. Can handle the problem at one time. But I know you have your idea. I have changed the code according to your requirements

joelmartinez:
@maopeixia ahh I can see that now ... with that understanding the code would have been fine as it was. It's also fine the way it is now that you've made these changes :) Thanks for accomodating the requests!

Just FYI, I'm not going to merge this PR just yet ... I'm preparing a release right now, so as soon as that goes out, I'll merge this and it will be ready for the follow up release. Thanks!

PR #498, opened 2020-05-09
monikerize interfaces and member implements in ECMAXML

TianqiZhang:


TianqiZhang:
@joelmartinez it seems integration tests failed due to this change, is it because I changed the schema?

TianqiZhang:
> @joelmartinez it seems integration tests failed due to this change, is it because I changed the schema?

Never mind, it turns out that I got the schema wrong. Fixed it.

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -2839,7 +2846,13 @@ private static void AddImplementedMembers(FrameworkTypeEntry typeEntry, MemberRe
             foreach (var implementedMember in implementedMembers)
             {
                 var value = msxdocxSlashdocFormatter.GetDeclaration(implementedMember);
-                WriteElementWithText(e, "InterfaceMember", value);
+                var interfaceMemberElement = WriteElementWithText(e, "InterfaceMember", value);
+
+                interfaceMemberElement.AddFrameworkToElement(typeEntry.Framework);
+                if (typeEntry.IsMemberOnLastFramework(mi))
```
joelmartinez, line 2852: Just like the check you added in `UpdateType` for `typeEntry.IsOnFirstFramework`, there should be a similar check here at the member level ... you can use `typeEntry.IsMemberOnFirstFramework(member)` to determine when you should clear the list of interface members (unless this is already happening somewhere and it's just not in these diffs?))

TianqiZhang: oh yes, the logic is already there at line 2808:
```
            var fingerprint = DocUtils.GetFingerprint(mi);
            if (typeEntry.IsMemberOnFirstFramework(mi))
            {
                ClearElement(root, "Implements");
            }
```

joelmartinez: excellent!

joelmartinez:
@TianqiZhang ... how do you want to coordinate this release going out ? Should we merge the return type changes and this to release as 5.7.6?
https://github.com/mono/api-doc-tools/pull/496

PR #500, opened 2020-05-27
[mdoc] come up with test/design for better member implementation dete…

maopeixia:
bug: https://ceapex.visualstudio.com/Engineering/_workitems/edit/180984
workitem  : https://ceapex.visualstudio.com/Engineering/_workitems/edit/219548
workitem  :https://ceapex.visualstudio.com/Engineering/_workitems/edit/219552

### DIFF
--- a/mdoc/mdoc.Test/MDocUpdaterTests.cs
+++ b/mdoc/mdoc.Test/MDocUpdaterTests.cs
@@ -101,5 +101,62 @@ namespace mdoc.Test
             //   foreach (var delitem in delList)
             // delitem.ParentNode.RemoveChild(child);        
         }
+
+        [Test]
+        public void UpdateToRight_MethodInterface_Test()
+        {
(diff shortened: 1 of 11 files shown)


mdoc/mdoc.Test/MDocUpdaterTests.cs
```diff
@@ -76,5 +77,59 @@ public void InternalEIITest()
...
+
+            var doc = new XmlDocument();
+            doc.LoadXml(XmlConsts.EiiErrorImplement);
+
+            var node = doc.SelectSingleNode($"/Type/Members/Member[@MemberName='{XmlNodeName}']");
+
+            if (node != null)
+            {
+                BindingFlags flags = BindingFlags.NonPublic | BindingFlags.Static;
+                MethodInfo mInfo = typeof(MDocUpdater).GetMethod("AddImplementedMembers", flags);
```
joelmartinez, line 124: actually ... this test looks great, but I'd rather prefer that you just make the `AddImplementedMembers` method public, and invoke it directly, rather than using reflection :)

PR #503, opened 2020-06-19
Support F# style Attribute rendering

TianqiZhang:


TianqiZhang:
@joelmartinez this PR is now ready for review. I reordered the commits so each one is meaningful and self-explanatory. Sorry for such long file list, 95% of them are integration test data 😄

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -3626,124 +3623,129 @@ static bool UpdateAssemblyVersionForAssemblyInfo (XmlElement e, XmlElement root,
...
-                    WriteElementText (ae, "AttributeName", currentAttribute.Key);
-
-                    if (fx.FrameworksCount > 1)
-                    {
-                        ae.SetAttribute (Consts.FrameworkAlternate, fx.Name);
-                    }
+                e = root.OwnerDocument.CreateElement("Attributes");
+
+            // we'll use C# version as key identifier of each attribute
+            var existingAttrElements = e.ChildNodes.SafeCast<XmlElement>()
```
joelmartinez, line 3666: Ok ... so it looks these changes were based on an older method for writing monikerized values (the attributes were an older implementation), which just tries to update the existing state, rather than what we've moved to that uses a common method to deal with the complexity of managing the FXA attributes. This common method basically clears elements the first time it comes across it, and rebuilds it every time mdoc is run, which I've found to be a cleaner and less bug-prone method.

You can see an example of this method's use here ... you basically just have to pass in some values and delegates (which lets you customize how it works for different cases) and then it takes care of everything else.
https://github.com/mono/api-doc-tools/blob/develop/mdoc/Mono.Documentation/MDocUpdater.cs#L4185-L4206

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -3626,124 +3623,129 @@ static bool UpdateAssemblyVersionForAssemblyInfo (XmlElement e, XmlElement root,
...
+                        else // let's create it
+                        {
+                            XmlElement ae = root.OwnerDocument.CreateElement("Attribute");
+                            e.AppendChild(ae);
+                            var node = WriteElementText(ae, "AttributeName", csharpAttrStr, forceNewElement: true);
+                            if (!isAssemblyAttribute)
+                            {
+                                WriteElementAttribute(node, Language, Consts.CSharp);
+
+                                foreach (var formatter in FormatterManager.AdditionalAttributeFormatters)
```
joelmartinez, line 3700: If you end up changing it per the previous comment, I would sort of invert this, where I iterate through all the relevant formatters (C# would be the default unless you added additional ones via -lang) ... and then just use that formatter's language (I assume there's a language attribute string?) to query for whether the attribute element exists.

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -3626,124 +3623,129 @@ static bool UpdateAssemblyVersionForAssemblyInfo (XmlElement e, XmlElement root,
...
+                        stateAttribute.Value.ParentNode.RemoveChild(stateAttribute.Value);
                 }
             }
 
             // clean up
-            bool lastFxForAssembly = (!string.IsNullOrWhiteSpace (assemblyName) && fx.IsLastFrameworkForAssembly (assemblyName));
-            bool lastFxForType = (typeEntry != null && fx.IsLastFrameworkForType (typeEntry));
+            bool lastFxForAssembly = (isAssemblyAttribute && fx.IsLastFrameworkForAssembly(assemblyName));
+            bool lastFxForType = (typeEntry != null && fx.IsLastFrameworkForType(typeEntry));
             if (lastFxForAssembly || lastFxForType)
```
joelmartinez, line 3744: In retrospect, this should definitely be clearer, because it might be prone to error ... this method should know whether we're adding assembly attributes, or type attributes (or really, any of the other attribute scenarios, like on parameters and return types)

TianqiZhang:
Thanks for the comments @joelmartinez , I've updated the PR to use `DocUtils.AddElementWithFx`. The logic is much shorter now 😄

TianqiZhang:
Here's the diff commit in dotnet-api-docs:
https://github.com/dotnet/dotnet-api-docs/commit/da20c4f8e76177855ffbac4eebcb2c8a23e09f0d

PR #512, opened 2020-08-05
method getter incorrectly marked as implementing nonexistant interfac…

maopeixia:
…e member


maopeixia:
work item: https://ceapex.visualstudio.com/Engineering/_workitems/edit/272318

maopeixia:
image
image
For the test of this bug, I tested it before made pr . Because the unit test is unable to imitate this effect(IsSpecialName = false),so simply changed the property  for testing.

PR #518, opened 2020-09-27
Duplicate node of non forwardingtype generated XML file

maopeixia:


joelmartinez:
In the end, we're going to solve this problem in another way. Thanks for the contribution :)

PR #524, opened 2020-11-30
add yaml files for azure pipeline

v-zazhou:


azure-pipelines.yml
```diff
@@ -0,0 +1,105 @@
...
+        }
+      ]
+
+- task: ArchiveFiles@2
+  displayName: Archive mdoc Files
+  inputs:
+    rootFolderOrFile: '$(mdocPath)'
+    includeRootFolder: false
+    archiveType: 'zip'
+    archiveFile: '$(Build.ArtifactStagingDirectory)/zips/mdoc-$(Build.BuildId).zip'
```
anmeng10101, line 97: for artifact version, shall we consider this format $(Year:yyyy).$(Month).$(DayOfMonth)$(Rev:.r) ?

azure-pipelines.yml
```diff
@@ -0,0 +1,105 @@
...
+    replaceExistingArchive: true
+
+- task: PublishBuildArtifacts@1
+  displayName: 'Publish mdoc Artifact'
+  condition: ne(variables['Build.Reason'],'PullRequest')
+  inputs:
+    PathtoPublish: '$(Build.ArtifactStagingDirectory)/zips'
```
anmeng10101, line 104: targetPath ?

hiihellox10:
I think `$(Year:yyyy)-$(Month)-$(DayOfMonth)-$(Rev:.r)` is better.

azure-pipelines.yml
```diff
@@ -0,0 +1,105 @@
...
+  
+steps:
+- task: NuGetToolInstaller@1
+  displayName: Install NuGet Tool
+
+- task: Bash@3
+  displayName: Run Unit and Integration Tests
+  inputs:
+    targetType: 'inline'
+    script: 'make prepare all check CONFIGURATION=Debug'
```
joelmartinez, line 24: We can just leave the `CONFIGURATION` parameter as the default since it defaults to `Release` ... however you could also pass in `$(buildconfiguration)` then that way it's configurable via the build.

azure-pipelines.yml
```diff
@@ -0,0 +1,105 @@
...
+    
+- task: MSBuild@1
```
joelmartinez, line 26: This task is unnecessary, as mdoc would have already been built in the previous bash task

azure-pipelines.yml
joelmartinez: Yeah the artifact name I don't think matters that match, but adding the date into the name would definitely make it informative :)

joelmartinez (changes requested):
This is looking great ... just a few small changes to streamline :)

joelmartinez (approved):
Thanks for the changes, this looks good :)

PR #526, opened 2020-12-07
Sort for assembly files and update expected XML files

hiihellox10:
The Directory.GetFiles method returned file names is not sort, this makes the order of the assembly elements of our generated XML files is inconsistent in different environments, so we need to sort it and update those expected XML files.


anmeng10101:
Hi @hiihellox10, let's update the commit message and pr description to help others understand why we need this code change.

joelmartinez (approved):
This looks great! 

@anmeng10101, I can just change the commit message when I do the squash commit :) And as for the changes on dotnet-api-docs, I think we can accept any changes that are just the result of a .dll re-ordering; this will make results much more consistent going forward so it's necessary.

Thanks @hiihellox10!

PR #532, opened 2021-01-15
Add support for C# 8.0 nullable reference types feature

hiihellox10:
The Roslyn team has written a good introduction about [Nullable Reference Types](https://github.com/dotnet/roslyn/blob/master/docs/features/nullable-reference-types.md) and [Nullable Metadata](https://github.com/dotnet/roslyn/blob/master/docs/features/nullable-metadata.md),  the articles explain how is the feature designed and implemented.

Nullable reference types metadata is similar to dynamic type metadata they both have the same implement, so I improve the DynamicParserContext class and rename DynamicParserContext to IAttributeParserContext. Currently, the IAttributeParserContext has already supported parse both types of metadata.


joelmartinez (approved):
👍🏼

PR #534, opened 2021-01-18
Add test project of nullable reference types

hiihellox10:
Add test project of nullable reference types and move the mdoc.Test.Cplusplus project to the mdoc.Test folder.


joelmartinez (approved):
This is awesome, thanks for moving the cplusplus project as well

PR #535, opened 2021-01-22
Naming unnamed parameter

hiihellox10:
Some auto-generation of API members does not have a name so I need to fix the issue, when an unnamed parameter is detected that it will be named unnamedParam[n + 1] now.


joelmartinez (approved):
This looks great, thank you!

PR #536, opened 2021-01-22
mdoc 5.8.1

joelmartinez:


hiihellox10:
@joelmartinez the return type issue is fixed.

mdoc/mdoc.Test/DotnetCoreAssemblyResolver.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        private void AddDotnetCoreToSearchDirectory()
+        {
+            if (Environment.OSVersion.Platform == PlatformID.Unix ||
+                Environment.OSVersion.Platform == PlatformID.MacOSX)
+            {
+                var latestDotnetCorePath = GetLatestPlatformAssembliesPath();
+                if (string.IsNullOrEmpty(latestDotnetCorePath))
+                {
+                    throw new DirectoryNotFoundException("The platform assemblies of .NET Core was not found, do you have .NET Core installed?");
```
joelmartinez, line 25: Please remove this exception ... you can replace it with a `Console.Write` with this message, however we should not be blocking mdoc from running, as it's entirely possible that a user could be using non-.net core set of assemblies, or they could have provided all the requested dependencies in a dependency folder, thus removing the need to resolve a system-installed version of .net core.

joelmartinez (commented):
@hiihellox10 These changes look good, but one small change please ... we generally try to be as permissive with mdoc's running parameters, to avoid hard stops for system configuration issues; opting instead to be verbose in the log with what issues are found.

mdoc/mdoc.Test/DotnetCoreAssemblyResolver.cs
hiihellox10: OK, but if the user doesn't have .NET Core installed an AssemblyResolutionException will be throwing when [a .NET Core type is resolving](https://github.com/mono/api-doc-tools/blob/0d14f8c4d3a7cd88eca1ee487652883ef74de3d2/mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs#L339). I think throwing a meaningful exception with a friendly message is better than the AssemblyResolutionException.

joelmartinez: @hiihellox10 I can completely understand your intent/reasoning here ... but know that over time we've come to learn that it's more desirable to lean towards a model where logging an error or warning is preferable to throwing an exception and potentially stopping the entire pipeline from running. 

In some cases, teams prefer to move ahead with incomplete results in the short term, and then fix them later ... in the case of throwing a type resolution error, yes that would block the process, but it would tell them precisely which type was unable to be resolved and the log would inform them of .net core not being installed. 

Additionally, they could explicitly prefer this in a scenario where they did not want to, or were unable to, install .net core on the build server ... but they would provide all dependencies in a dependencies folder. This would allow their build to complete without requiring a change to the build agent.

PR #537, opened 2021-01-27
Fix signature issue of the .NET Core type

hiihellox10:
The Mono.Cecil can't load .NET Core platform assemblies in WSL, macOS, and Ubuntu OS environment that an AssemblyResolutionException will be throwing when a .NET Core type be resolve but it can work in the Windows OS environment as well.

Thus the unit tests of nullable reference type can't work in the Linux OS environment but it can work in the Windows OS environment as well. I did [some work](https://github.com/mono/api-doc-tools/blob/develop/mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs#L337) so that it could work in the Linux OS environment but also introduce the new bug into mdoc.

I found another way to load .NET Core platform assemblies for the unit tests. We will load .NET Core assemblies in the external\Test folder by a custom IAssemblyResolver. But mdoc still can't analyze and generate .NET Core binaries signature in Linux OS environment, I don't fix the issue now. Our pipeline of docs reference is work in windows looks like we don't have that requirement.


joelmartinez:
Hey Xiao ... thanks for this contribution, however I'm going to close the PR because we are actively going to avoid adding platform assemblies to this repository. If we were to approach properly fixing this, we would want to adjust the logic in this method in the resolver:
https://github.com/mono/api-doc-tools/blob/master/mdoc/Mono.Documentation/Updater/Frameworks/MDocResolver.cs#L512-L529

Giving it the ability to find those system assemblies on its own

PR #539, opened 2021-03-11
Improve argument value formatter that fix the value of attribute argument not imported correctly

hiihellox10:
When a defined property of an attribute is an object type that the value of the property cannot be correctly identified. For example, the [ConfigurationPropertyAttribute.DefaultValue](https://docs.microsoft.com/en-us/dotnet/api/system.configuration.configurationpropertyattribute.defaultvalue?view=net-5.0) property is an object type so the value of the DefaultValue property can be of any type but we don't support convert the value to a correct type before. I create some test cases to reproduce and fix the issue of the same type. Now we have supported all the [attribute parameter types](https://github.com/dotnet/csharplang/blob/main/spec/attributes.md#attribute-parameter-types).


mdoc/Mono.Documentation/Updater/Formatters/AttributeFormatters/AttributeValueFormatter.cs
```diff
@@ -1,15 +1,20 @@
...
 namespace Mono.Documentation.Updater
 {
     /// <summary>Formats attribute values. Should return true if it is able to format the value.</summary>
-    class AttributeValueFormatter
+    public class AttributeValueFormatter
     {
+        private readonly string[] numbersTypeFullNames =
+        { 
+            "System.Int32",
+            "System.Int64"
```
joelmartinez, line 15: 🤔 there's a few more number types than this, for example: Int16 (short), Single (float), Double, and Decimal

mdoc/Mono.Documentation/Util/AssemblyLoader.cs
```diff
@@ -0,0 +1,607 @@
+﻿using Mono.Cecil;
```
joelmartinez, line 1: There's something strange about the changes in this file ... can you revert changes on AssemblyLoader to avoid adding unknown change noise?

mdoc/mdoc.Test/AttributeFormatterTest.cs
```diff
@@ -0,0 +1,90 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace mdoc.Test
+{
+    [TestFixture()]
+    public class AttributeFormatterTest : BasicTests
+    {
+        [TestCase("PropertiyObjectWithBoolType", true)]
```
joelmartinez, line 15: Misspelled "Property" as "Propertiy"

joelmartinez (changes requested):
Thanks for opening the PR ... I've made a few comments. And also looking at the build result, it seems to be failing on a strange diff; doesn't look like there's any actual change here?

image

In any case, please let us know when you've got this working, or otherwise if you come across any other blockers. Thanks!

mdoc/mdoc.Test/AttributeFormatterTest.cs
hiihellox10: Thanks!

mdoc/Mono.Documentation/Updater/Formatters/AttributeFormatters/AttributeValueFormatter.cs
hiihellox10: We don't need those numeric types, they just for tests only. I will remove them later.

mdoc/Mono.Documentation/Util/AssemblyLoader.cs
hiihellox10: This class AssemblyLoader just for testing some special enum type only, I will remove it later. @anmeng10101 said maybe you want to know what I did lately so I open this draft PR for you but the issue has not been fixed.

joelmartinez:
@hiihellox10 I see the tests are passing now, which is great! Is the PR ready for review? if so, please publish it (it's currently a draft)

hiihellox10:
> @hiihellox10 I see the tests are passing now, which is great! Is the PR ready for review? if so, please publish it (it's currently a draft)

@joelmartinez the issue is closer to fixed but it is have not complete that still have a few related issues need to be resolved.

joelmartinez (approved):
Thanks for the contribution, this will really improve attribute value rendering

PR #540, opened 2021-03-19
Copy typemap.xml to output

jeffreyxrh:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/394298/


joelmartinez (approved):
This looks great, thank you. I'm not going to complete it immediately as we're currently working on completing another mdoc release prior to the one this one will go out in, but this work looks good.

PR #543, opened 2021-04-02
Added missing const& for cpp/winrt syntax

jeffreyxrh:
Link to work item: https://dev.azure.com/ceapex/Engineering/_workitems/edit/86667
This change will add '&' for out parameters and 'const &' for in parameters for all Cpp/WinRt parameter syntax.


joelmartinez (changes requested):
Looks like this change has introduced some changes that are causing failures on the unit test suite. You will likely have to make some changes to the expected test results. Take a look at the contribution guide to learn a bit more about our test suite, and in particular the difference between the unit tests and integration tests:
https://github.com/mono/api-doc-tools/blob/main/CONTRIBUTING.md#tests

jeffreyxrh:
@joelmartinez Updated integration tests and now they all pass. Please verify the changes. Additionally, removed space between the `const` and the `&`'.

joelmartinez (approved):
Thanks @jeffreyxrh!

PR #548, opened 2021-04-19
Added type forwarding check

jeffreyxrh:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/373824


jeffreyxrh:
https://github.com/Azure/azure-docs-sdk-dotnet/compare/master...jeffreyxrh:jex/373824 diff report generated 20210416

mdoc/Mono.Documentation/MDocUpdater.cs
```diff
@@ -3871,9 +3871,22 @@ static bool UpdateAssemblyVersionForAssemblyInfo (XmlElement e, XmlElement root,
...
+        private bool ProcessedMoreThanOnce(FrameworkTypeEntry typeEntry)
+        {
+            if (typeEntry.TimesProcessed <= 1)
+            {
+                return false;
+            }
+            else
+            {
+                var assemblies = this.assemblies.Where(a => a.Name == typeEntry.Framework.Name).ToList();
+                return assemblies.Any(a => a.IsTypeForwardingTo(typeEntry));
```
joelmartinez, line 3883: There are two similar but independent scenarios that this fix should be addressing:

1. A type forward assembly +  the assembly with the actual definition
2. Two separate assemblies, with the same type (name), but different `TypeDefinition`s (ie. different members ... such as different extension methods depending on which assembly reference you've added).

I would like to see some integration tests that reflect both of these scenarios. In fact, there's already a [type forward test included here](https://github.com/mono/api-doc-tools/blob/main/mdoc/Makefile#L203) ... maybe you could add to that [particular test](https://github.com/mono/api-doc-tools/blob/main/mdoc/Test/DocTest-typeForwards.cs), a separate test class that reflects the second scenario, to make sure that mdoc will correctly add all methods in both classes.
```suggestion
                return assemblies.Any(a => a.IsTypeForwardingTo(typeEntry));
```

joelmartinez (changes requested):

PR #558, opened 2021-05-18
mdoc should not use TypeMap for parameter and return elements

michaeljin89757:


joelmartinez (changes requested):
This is great, but the same type of change also needs to be added for parameter types. Look at the `MakeParameters` function

joelmartinez (changes requested):
Thank you, yes this change looks good as well, and I appreciate that you added the default parameter so that any other callers who might be using `GetDocParameterType` will still get the original behavior.

In my haste yesterday, I asked for only this one change ... but that was partially because I was reacting to an urgent need to unblock @TianqiZhang. In the meantime, @anmeng10101 has created a preview release that includes this change so that has unblocked. So I would like to request one other change.

It's interesting that this hasn't triggered any changes in our expected integration test files ... which means that we must not have any typemap integration tests. I'd like to request that you add an integration test, along with expected XML files, that uses a typemap file, and we see the mapped type in the signature, but the original type in the return and parameter elements.

This way we can codify the behavior in our integration test suite, and better guard against regressions in the future. Thanks!

michaeljin89757:
Parameter
ReturnValue
 Here are two pictures of before and after change.
 On the left side is the result after the change. On the right side is the result from a mdoc local run using main branch.

mdoc/Makefile
```diff
@@ -852,6 +857,22 @@ check-ignored-namespace-generic:
...
+check-type-projection: Test/DocTest-TypeProjection.dll
+	rm -Rf Test/en.actual
+	rm -Rf Test/test-type-projection
+
+	mkdir Test/test-type-projection
+	mkdir Test/test-type-projection/DocTest-TypeProjection
+	mkdir Test/en.actual
+
+	cp Test/DocTest-TypeProjection.dll Test/test-type-projection/DocTest-TypeProjection
+	cp Test/TypeMap.xml Test/test-type-projection/
```
joelmartinez, line 870: You should give `TypeMap.xml` a different name in the `Test` directory ... something that makes it clearer which test it's related to. You can rename it after you copy with the `mv` command.

mdoc/Test/en.expected-fsharp-wsl/Delegates+Delegate10.xml
```diff
@@ -1,5 +1,5 @@
 <Type Name="Delegates+Delegate10" FullName="Delegates+Delegate10">
-  <TypeSignature Language="C#" Value="public delegate char Delegates.Delegate10(int , int );" />
+  <TypeSignature Language="C#" Value="public delegate char Delegates.Delegate10(int unnamedParam1, int unnamedParam2);" />
```
joelmartinez, line 2: It's really strange that we're getting all these changes only now in this PR, when the unnamed parameter change happened all the way back in february with #535 🤔

mdoc/Test/en.expected-type-projection/Mono.DocTest/IList.xml
```diff
@@ -0,0 +1,38 @@
...
+      <MemberType>Method</MemberType>
+      <AssemblyInfo>
+        <AssemblyName>DocTest-TypeProjection</AssemblyName>
+        <AssemblyVersion>0.0.0.0</AssemblyVersion>
+      </AssemblyInfo>
+      <ReturnValue>
+        <ReturnType>Mono.DocTest.IVector</ReturnType>
+      </ReturnValue>
+      <Parameters>
+        <Parameter Name="A" Type="System.Collections.IList" />
```
joelmartinez, line 26: I would have expected this parameter type to be `Mono.DocTest.IVector` as well just like in the return value ... this result here shows that it's still using the type map

joelmartinez (changes requested):

joelmartinez (approved):
Thanks for the great work on adding the integration testing ... really helped capture some things that otherwise would have gone unnoticed. This LGTM /cc @anmeng10101

PR #565, opened 2021-06-30
Bug 388663: update ECMAUrlParser to support nullable (?)

TianqiZhang:


monodoc/Monodoc.Ecma/EcmaDesc.cs
```diff
@@ -27,7 +27,8 @@ public enum Mod
 			Normal,
 			Pointer,
 			Ref,
-			Out
+			Out,
+			Nullable
```
joelmartinez, line 31: Make sure you add unit tests in the `[EcmaUrlTests](https://github.com/mono/api-doc-tools/blob/24ba59fcca3d1acacdb0f71acbdf401d5ce3fb96/monodoc/Test/Monodoc.Ecma/EcmaUrlTests.cs)https://github.com/mono/api-doc-tools/blob/24ba59fcca3d1acacdb0f71acbdf401d5ce3fb96/monodoc/Test/Monodoc.Ecma/EcmaUrlTests.cs` file

TianqiZhang: done

joelmartinez (approved):
LGTM!

PR #576, opened 2021-09-18
Fixed the issue of dual property const

JeffInChrist:
The "System.Windows.Media.Animation.Storyboard.TargetProperty" attached property is missing from the .NET API ref docs. I am adding it back now. Besides, it is just a initial version and this attached property needs authoring.

```
<MemberSignature Language="DocId" Value="F:System.Windows.Media.Animation.Storyboard.TargetProperty" />
<MemberSignature Language="DocId" Value="P:System.Windows.Media.Animation.Storyboard.TargetProperty" />
```
Related Work Item:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/283041

### DIFF
--- a/mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs
+++ b/mdoc/Mono.Documentation/Util/AttachedEntitiesHelper.cs
@@ -130,12 +130,14 @@ namespace Mono.Documentation.Util
             // https://github.com/mono/api-doc-tools/issues/63#issuecomment-328995418
             if (!field.Name.EndsWith(PropertyConst, StringComparison.Ordinal))
                 return false;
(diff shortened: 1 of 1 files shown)


joelmartinez (changes requested):
This is a pretty unique situation, so I would say that we should probably have a test here to track this functionality for future revisions (either unit or integration)

JeffInChrist:
@joelmartinez Good idea. I've added several tests for these changes. Additionally, it seems the attached entities should be enriched for UPW and WinUI. So I've made a bit changes, too.

joelmartinez (approved):
Thanks for adding the additional testing 👏🏼 lgtm

PR #586, opened 2021-11-08
Removed types of exteranl assemblies for forwarded types.

JeffInChrist:
It is intended for dotnet platform extensions. There are redundant types from dependency assemblies, resulting in OPS build warnings for invalid assembly monikers. The changes can eliminate these types in framework alternates.

Related Work Item:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/491194


huangmin-ms:
Not sure why we need to union exported types here.
https://github.com/mono/api-doc-tools/blob/main/mdoc/Mono.Documentation/Util/CecilExtensions.cs#:~:text=var%20exportedTypes%20%3D%20assembly,GetAllTypes%20()).Union(exportedTypes)%3B

I think **NOT** union the exported types will solve this issue as well.

huangmin-ms:
Close this PR as we have an alternate PR https://github.com/mono/api-doc-tools/pull/592

PR #588, opened 2021-11-22
Fix bug 514755

v-susu1:
Fix bug 514755 that handle value type with nullable tag


mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs
```diff
@@ -390,13 +390,32 @@ private bool IsTypeOfOperatorWithUnboundGenericType(TypeReference type, IEnumera
             return false;
         }
 
-        private void AppendGenericTypeParameter (StringBuilder buf, TypeReference type, IAttributeParserContext context, bool useTypeProjection = true)
+        private bool IsGenericWithValueTypeArg(GenericInstanceType type)
```
huangmin-ms, line 393: does isNullableValueType sound better?

mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs
```diff
@@ -390,13 +390,32 @@ private bool IsTypeOfOperatorWithUnboundGenericType(TypeReference type, IEnumera
...
+        {
+            if (type.Name == "Nullable`1" && type.HasGenericArguments && type.GenericArguments[0].IsValueType)
```
huangmin-ms, line 395: please help to check if we can directly use  **type.IsValueType**  instead of **type.HasGenericArguments && type.GenericArguments[0].IsValueType**

mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs
```diff
@@ -390,13 +390,32 @@ private bool IsTypeOfOperatorWithUnboundGenericType(TypeReference type, IEnumera
...
+
+        private void AppendGenericTypeParameter(StringBuilder buf, TypeReference type, IAttributeParserContext context, bool useTypeProjection = true)
         {
             var isNullableType = false;
             if (type is GenericInstanceType)
             {
-                isNullableType = context.IsNullable();
-                buf.Append(GetTypeName(type, context));
+                // Fix bug https://dev.azure.com/ceapex/Engineering/_workitems/edit/514755?src=WorkItemMention&src-action=artifact_link
+                if (IsGenericWithValueTypeArg(type as GenericInstanceType))
```
huangmin-ms, line 409: if (type is GenericInstanceType)
{
  isNullableType = isNullableValueType(type as GenericInstanceType) ? true : context.IsNullable();
  buf.Append(GetTypeName(type, context));
}

huangmin-ms:
please help to add cases to cover:
<string?, string?, bool?>
<string?, bool?, string?>
<string?, bool?, int?, string?>
<bool?>

mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs
v-susu1: updated.

mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs
v-susu1: Since **public struct Nullable<T> where T : struct** , so we can just only check "type.Name == "Nullable`1"

mdoc/Mono.Documentation/Updater/Formatters/MemberFormatter.cs
v-susu1: updated.

huangmin-ms (approved):

PR #608, opened 2022-02-15
feat#550375: Support ref struct and readonly ref struct syntaxes

huangmin-ms:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/550375

**ref struct**
before: https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/ArgIterator.xml
after: 
image

**readonly ref struct**
before: https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/Span%601.xml
after:
image

**readonly struct**
before: https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Runtime.Intrinsics/Vector64%601.xml
after:
image

### DIFF
--- a/mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
+++ b/mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
@@ -238,7 +238,19 @@ namespace Mono.Documentation.Updater.Formatters
             if (t.IsEnum)
                 return "enum";
             if (t.IsValueType)
-                return "struct";
+            {
+                StringBuilder buf = new StringBuilder();
+                if (DocUtils.HasReadOnlyAttribute(t))
+                {
+                    buf.Append("readonly ");
+                }
+                if (t.CustomAttributes.Any(attr => attr.AttributeType.FullName == "System.Runtime.CompilerServices.IsByRefLikeAttribute"))
+                {
+                    buf.Append("ref ");
+                }
+                buf.Append("struct");
+                return buf.ToString();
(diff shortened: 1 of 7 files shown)


mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
```diff
@@ -238,7 +238,19 @@ static string GetTypeKind (TypeDefinition t)
...
                 return "enum";
             if (t.IsValueType)
-                return "struct";
+            {
+                StringBuilder buf = new StringBuilder();
+                if (DocUtils.HasReadOnlyAttribute(t))
+                {
+                    buf.Append("readonly ");
+                }
+                if (t.CustomAttributes.Any(attr => attr.AttributeType.FullName == "System.Runtime.CompilerServices.IsByRefLikeAttribute"))
```
yishengjin1413, line 247: Shall we use a method the same as HasReadOnlyAttribute?

huangmin-ms: @yishengjin1413  We can reuse the ReadOnlyAttribute so I extract it to a method. But for this attribute, it only stands for `ref struct`. We are not able to reuse it elsewhere. So I think there's no need to extract it to a method.

huangmin-ms: @yishengjin1413  updated

yishengjin1413 (approved):

PR #620, opened 2022-03-14
feat#550401:Support C# 9 nint/nuint

huangmin-ms:
diffs:
https://github.com/dotnet/dotnet-api-docs/compare/3d94962...b358faa

feature link: https://ceapex.visualstudio.com/Engineering/_workitems/edit/550401


mdoc/Mono.Documentation/Updater/AttributeParserContext.cs
```diff
@@ -99,16 +107,36 @@ private void ReadNullableAttribute()
...
+        private void ReadNativeIntegerAttribute()
+        {
+            nativeIntegerFlags = ReadCustomAttributeValue<bool>(
+                () => new bool[] { true },
+                Consts.NativeIntegerAttribute);
+        }
+
+        private T[] ReadCustomAttributeValue<T>(Func<T[]> init, string attributeName)
         {
             if (provider != null && provider.HasCustomAttributes)
```
yishengjin1413, line 125: if (provider == null || !provider.HasCustomAttributes) return null; ?

How about return null first instead of a big if code block?

mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
```diff
@@ -56,6 +56,18 @@ protected virtual string GetCSharpType (string t)
                 case "System.Void": typeToCompare = "void"; break;
                 case "System.String": typeToCompare = "string"; break;
                 case "System.Object": typeToCompare = "object"; break;
+                case "System.IntPtr":
+                    if (context != null && context.IsNativeInteger())
```
yishengjin1413, line 60: Can we reduce these dupilcated code? And also shall we remove the context parameter here, this method looks a pure string mapping before, shall we keep it sample?

mdoc/Mono.Documentation/Updater/AttributeParserContext.cs
```diff
@@ -82,6 +85,11 @@ public string GetTupleElementName()
             return (tupleElementNames == null || tupleNameAttributeIndex >= tupleElementNames.Length) ? null : tupleElementNames[tupleNameAttributeIndex++];
         }
 
+        public bool IsNativeInteger()
+        {
+            return (nativeIntegerFlags == null || nativeIntegerAttributeIndex >= nativeIntegerFlags.Length) ? false : nativeIntegerFlags[nativeIntegerAttributeIndex++];
```
yishengjin1413, line 90: Is "return nativeIntegerFlags != null && nativeIntegerAttributeIndex < nativeIntegerFlags.Length) && nativeIntegerFlags[nativeIntegerAttributeIndex++]; " better?

huangmin-ms: Updated

mdoc/Mono.Documentation/Updater/AttributeParserContext.cs
huangmin-ms: updated

yishengjin1413 (approved):

PR #637, opened 2022-06-22
[MDocUpdater] Fix usage of the --delete option

pjcollins:
[MDocUpdater] Fix usage of the --delete option

Commit eb7fec62c introduced a breaking change that caused doc removal to
be skipped for members that were removed from a type:

    mdoc: Member Removed: File='/Users/peter/source/android-api-docs/docs/Mono.Android/en/Android.Text/AlteredCharSequence.xml'; Signature='System.Collections.IEnumerator IEnumerable.GetEnumerator ();'
    mdoc: Not deleting 'System.Collections.IEnumerator IEnumerable.GetEnumerator ();'; must be enabled with the --delete option

The error message is now misleading in this case.  I am using the
`--delete` parameter when running the tool, but it no longer works for
removed members.

To help address the scenario fixed in eb7fec62c, a new `--retain-eii`
parameter has been added to mdoc.  When this is parameter is set and
explicit interface implementation docs are not empty, they will not be
removed (even when `--delete` is used).

The `--delete` behavior has otherwise been reverted to its previous
state, and the new `--retain-eii` param can be set to opt in to this new
EII specific beahvior.

### DIFF
--- a/mdoc/Mono.Documentation/MDocUpdater.cs
+++ b/mdoc/Mono.Documentation/MDocUpdater.cs
@@ -34,6 +34,7 @@ namespace Mono.Documentation
         bool isClassicRun;
         bool verbose;
         bool delete;
+        bool retainEII;
         bool show_exceptions;
         bool no_assembly_versions, ignore_missing_types;
         ExceptionLocations? exceptions;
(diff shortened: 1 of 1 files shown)


pjcollins:
FYI @michaeljin89757 and @huangmin-ms for review, I don't have permission to add reviewers explicitly.

huangmin-ms:
@pjcollins The intent is to make sure even with `--delete` flag, the type/member which has user manually edited docs should not be deleted. if deleted, user doc will be lost and cannot be restored.

pjcollins:
@huangmin-ms When you have a chance to review I've reworked this to still preserve previous "delete" behavior, while adding a flag to try to address your concern with explicit interface implementations.

huangmin-ms:
@pjcollins  this issue will be addressed in #647

PR #638, opened 2022-06-22
[MDocUpdater] Fix assembly info duplication

pjcollins:
The `AssemblyPublicKey` and `AssemblyCulture` elements will be added to
to a docs `index.xml` file every time the tool runs, even if the element
already exists.  Fix this by using `MDocUpdater.WriteElement` to write
these elements, as it will first check for an existing element rather
than always creating a new one.


huangmin-ms:
@pjcollins  please target to `develop` branch for contribution. Thanks!

huangmin-ms (approved):

PR #639, opened 2022-06-23
[mdoc] Multitarget net471 and net6.0

pjcollins:
Adds `net6.0` as a target framework to mdoc.  The artifacts from the
`net6.0` build will be output to `bin/$(Configuration)-net6.0`.

Mono.Cecil has been bumped from `0.10.0-beta5` to `0.10.0`.  This update
appears to have introduced a change in behavior in F# doc generation,
and the expected output in `PatternMatchingExamples.xml` has been
updated to reflect that.

The mdoc NuGet package has been updated to include both `net471` and
`net6.0` tools.  This can be considered a breaking change, as the path
of mdoc.exe has been changed to include a folder named after its target
framework.

### DIFF
--- a/mdoc/Makefile
+++ b/mdoc/Makefile
@@ -61,6 +61,9 @@ cleanup:
 nunit: 
 	mono ../packages/NUnit.ConsoleRunner.3.6.0/tools/nunit3-console.exe mdoc.Test/bin/$(CONFIGURATION)/mdoc.Test.dll
 
+dotnet-test:
+	dotnet test mdoc.Test/bin/$(CONFIGURATION)-net6.0/mdoc.Test.dll
+
 Test/DocTest-VB-Eii.dll:
 	$(VBCOMPILE) -out:Test/DocTest-VB-Eii.dll  Test/ClassEnumerator.vb
 	
@@ -960,5 +963,5 @@ check-doc-tools-update: check-monodocer-since-update \
 	check-mdoc-export-msxdoc-update \
 	check-mdoc-validate-update 
 
-check: nunit check-doc-tools
+check: nunit dotnet-test check-doc-tools
 	@echo "mdoc Tests Complete!"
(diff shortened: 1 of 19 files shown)


huangmin-ms:
Thanks @pjcollins would you mind also upgrading mdoc.test.csproj to net6？

pjcollins:
@huangmin-ms I think it makes sense to at least have _some_ tests running against the net6.0 version of mdoc, we gain that new test coverage by running the net6.0 version of mdoc.Test.dll. Maybe we could file an issue that mentions the tests ignored for net6.0 to see if they can be re-enabled in the future.

mdoc/Makefile
```diff
@@ -960,5 +963,5 @@ check-doc-tools-update: check-monodocer-since-update \
 	check-mdoc-export-msxdoc-update \
 	check-mdoc-validate-update 
 
-check: nunit check-doc-tools
+check: nunit dotnet-test check-doc-tools
```
huangmin-ms, line 966: some cases will be run twice (net471 and net 6.0 test dll), right?

mdoc/mdoc.Test/mdoc.Test.csproj
```diff
@@ -37,10 +37,14 @@
       <HintPath>..\..\external\Windows\Windows.Foundation.UniversalApiContract.winmd</HintPath>
     </Reference>
   </ItemGroup>
+
   <ItemGroup>
     <PackageReference Include="Mono.Cecil" Version="$(NuGetVersionMonoCecil)" />
     <PackageReference Include="NUnit" Version="$(NuGetVersionNUnit)" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
```
huangmin-ms, line 44: should we define the version string in `Directory.Build.props`?

mdoc/Makefile
pjcollins: Yes most tests will run twice, but this is desirable as they will be running against different framework versions.

mdoc/mdoc.Test/mdoc.Test.csproj
pjcollins: I moved these two package versions to Directory.Build.props.

huangmin-ms (approved):

huangmin-ms:
@pjcollins Thanks for your contribution! Really appreciate it!

PR #641, opened 2022-06-24
[DocumentationEnumerator] Support global alias matching

pjcollins:
The tooling is currently unable to match existing documentation with a
member that was updated to include a global alias.  Fix this by adding a
a type lookup attempt that prefixes `global::` to the documentation
member name.

### DIFF
--- a/mdoc/Mono.Documentation/Updater/DocumentationEnumerator.cs
+++ b/mdoc/Mono.Documentation/Updater/DocumentationEnumerator.cs
@@ -256,8 +256,8 @@ namespace Mono.Documentation.Updater
             //  1. "Normal" (non-generic) member names: GetEnumerator
             //    - Lookup as-is.
             //  2. Explicitly-implemented interface member names: System.Collections.IEnumerable.Current
-            //    - try as-is, and try type.member (due to "kludge" for property
-            //      support.
+            //    - try as-is, try global::namespace.type.member, and try
+            //      type.member (due to "kludge" for property support.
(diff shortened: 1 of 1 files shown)


huangmin-ms (approved):

huangmin-ms:
@pjcollins  @jonpryor  please let me know if a new nuget package release is needed for you to consume this change.

pjcollins:
@huangmin-ms thanks! We will need a new NuGet package release with these changes, however I have a couple of other PRs that would be great to bundle with the next NuGet release if possible:

* https://github.com/mono/api-doc-tools/pull/637 - I've just tried to update this to address your comment about explicit interface implementations. However, I am not familiar with the scenario that commit eb7fec62c3ec4759ae178bd601564aa4f4d66ed6 was trying to fix and am not sure if I am entirely on the right track there.
* https://github.com/mono/api-doc-tools/pull/638 - Should hopefully be a straightforward change.

huangmin-ms:
@pjcollins  #638 is merged.  #642 may also be able to catch this release. For #637 is this one urgent? Can we postpone to next release? I may be able to look at this one sometime next week.

pjcollins:
@huangmin-ms Thanks! #637 is not immediately urgent, it can wait until the next release. It may be required for us when we try to update our Android API docs from Android 12 to Android 12.1 or Android 13, which we would like to do in the coming months.

huangmin-ms:
@pjcollins [mdoc 5.8.9.2](https://www.nuget.org/packages/mdoc/5.8.9.2) is released.

PR #662, opened 2023-01-17
[update] Try to keep existing order for type and member signatures.

rolfbjarne:
This avoids a lot of unnecessary churn in the diff when updating docs.


rolfbjarne:
@huangmin-ms I see you changed the target branch, do I need to do anything?

huangmin-ms:
@rolfbjarne  Not for now. I need to take a closer look at this PR and will let you know then.  The process is to first merge `develop` and then `develop` to `main`.

PR #711, opened 2024-09-14
C# signatures for new operators

nickwalkmsft:
Work items:

- https://ceapex.visualstudio.com/Engineering/_workitems/edit/722484
- https://ceapex.visualstudio.com/Engineering/_workitems/edit/722491

Implement correct C# method signature output for the newer operators that currently aren't implemented: unsigned right shift and all of the checked operators.

Added unit tests and tested locally against the types mentioned in the work items (BigInteger and Int128).

As with #710, I was unable to get C++ tests running, but nothing in that codepath should be affected.


huangmin-ms:
@nickwalkmsft  Thanks Nick. I granted you write permission to this repo, could you accept the invitation and directly push to a branch and then create a PR? Because the forked PR won't trigger the CI pipeline due to the security reason. And also the target branch should be `develop`.

nickwalkmsft:
Obsoleted by #713.

PR #712, opened 2024-09-25
#945509 Support ref readonly parameters

nickwalkmsft:
Work item: https://ceapex.visualstudio.com/Engineering/_workitems/edit/945509

Add support for correct rendering of "ref readonly" parameter in C# method signature documentation.

Added one unit test and also verified output against the case mentioned in the work item (https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe.asref?view=net-8.0#system-runtime-compilerservices-unsafe-asref-1%28-0@%29 should render with `scoped ref readonly T source` instead of `scoped ref T source`.

I couldn't get the C++ tests running but they appear to be out of scope here.

PR #713, opened 2024-09-25
C# signatures for new operators

nickwalkmsft:
Work items:

- https://ceapex.visualstudio.com/Engineering/_workitems/edit/722484
- https://ceapex.visualstudio.com/Engineering/_workitems/edit/722491

Implement correct C# method signature output for the newer operators that currently aren't implemented: unsigned right shift and all of the checked operators.

Added unit tests and tested locally against the types mentioned in the work items (BigInteger and Int128).

As with #710, I was unable to get C++ tests running, but nothing in that codepath should be affected.

### DIFF
--- a/mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
+++ b/mdoc/Mono.Documentation/Updater/Formatters/CSharpFullMemberFormatter.cs
@@ -392,6 +392,9 @@ namespace Mono.Documentation.Updater.Formatters
                 methodName = ifaceMethod.Name;
             }
 
+            // TODO this is liable to register false positives, but that's only likely to happen
+            // in an adversarial situation. Operators have to be public static, with the
+            // specialname bit set, with the right parameters for the type of operation.
             if (methodName.StartsWith("op_", StringComparison.Ordinal))
             {
                 // this is an operator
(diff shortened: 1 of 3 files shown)


v-fuquanli:
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=1868434

huangmin-ms:
> https://ceapex.visualstudio.com/Engineering/_build/results?buildId=1868434

@nickwalkmsft  The automatically triggers stop working somehow. As a workaround, you can run it manually in https://ceapex.visualstudio.com/Engineering/_build/results?buildId=1868434&view=results and enter your branch name.
image
Please make sure the pipeline is passed.

nickwalkmsft:
@huangmin-ms bummer, both this and #712 failed on compilation errors. It looks like the build uses .NET6/C#10, but checked user-defined operators weren't added until .NET7/C#11, and ref readonly parameters in .NET8/C#12.

What does it take to update the platform version used to compile and run mdoc? I imagine it has the potential to require some breakfixes, updates to the runtime environments used in the .NET pipelines that run mdoc, etc.

nickwalkmsft:
Thanks for digging on this @huangmin-ms! I admittedly didn't spend much time looking at the build, I just set up a mostly-working dev environment using VS2022 locally and worked on the code from there, so it doesn't surprise me that there are build-related problems.

There's no hurry for these items, they can remain at their current priority, so please don't feel there's any rush to resolve the issues. Thanks!

PR #720, opened 2024-12-18
Fix bug : Cross reference not found: 'System.IDisposable.Close'.

v-fuquanli:
**Fix bug : Cross reference not found: 'System.IDisposable.Close'.**
WorkItem : https://ceapex.visualstudio.com/Engineering/_workitems/edit/1018526

**Before the problem was fixed**
image
image

**After fixing the problem**
System.IDisposable.Close disappeared, and the rest of the xref-not-found were actually only about 40 or so, because the count was by word, and there were multiple identical words in a row.
image

### DIFF
--- a/mdoc/Consts.cs
+++ b/mdoc/Consts.cs
@@ -3,7 +3,7 @@ namespace Mono.Documentation
 {
     public static class Consts
     {
-        public static string MonoVersion = "5.9.3.6";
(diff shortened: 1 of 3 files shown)


v-fuquanli:
https://dev.azure.com/ceapex/Engineering/_build/results?buildId=1943985&view=results

huangmin-ms (approved):

PR #737, opened 2025-10-30
Support ref readonly parameters (#712)

bianyifan:
Add support for correct rendering of "ref readonly" parameter in C# method signature documentation.


xmdanni:
https://dev.azure.com/ceapex/Engineering/_build/results?buildId=2295734

PR #741, opened 2025-11-03
Support .NET 10 extension members

huangmin-ms:

PR #742, opened 2025-11-05
Support .NET 10 extension members (#741)

huangmin-ms:
* Support .NET 10 extension members

* Fix typename

* Update

* bump version to 5.9.6


xmdanni:
https://dev.azure.com/ceapex/Engineering/_build/results?buildId=2302973

PR #744, opened 2026-04-24
Fix Recursion for C# Formatter for ValueTuple

davidle-ms:
Resolves [#11939](https://github.com/dotnet/dotnet-api-docs/issues/11939).

Summary: 
- Fixed C# Formatter to correctly address System.ValueTuple nested ValueTuple structure, where there are more than 7 required fields, it takes the recursive structure: System.ValueTuple<T1, ..., T7, TRest>
- Added unit test cases to validate this change
- Tested this locally with System.ValueTuple.dll, comparing the generated xmls before and after the change. Change in xml file is as expected.

PR #745, opened 2026-04-27
Fix ValueTuple Formatting Error for C# Formatter (#744)

davidle-ms:
* fix: recursion with c# formatter for valuetuple

* docs: add comment to clearly explain change

* test: add test cases for method signature for valuetuple

* chore: fix unnecessary loop change

* chore: update version to 5.9.7


xmdanni:
https://dev.azure.com/ceapex/Engineering/_build/results?buildId=2521433

